PR #2, opened 2020-06-11
RFC: AQL-Based Table Creation

waxlamp:
This RFC describes a proposal for adding API to create new tables from AQL
queries. This supercedes a [rejected RFC for a View API](view_api_rejected.md),
offering a simpler approach.

The proposal is to add a way to run an AQL query as the basis for data to
"upload" to a table, fulfilling the target use cases of the View API idea, and
providing an eventual basis for [virtual tables](https://docs.google.com/document/d/15WMH_FO10qDPtwu5A_uq4Ex2tWoOcVZsGmt1e-nNM3Q/edit)
in Multinet.

This RFC is derived ultimately from a [Google Doc proposal](https://docs.google.com/document/d/1C8VDA867jn6sLnLtOlsZAJVn7nGxg5lGq7OSYcPa848/edit).


rfcs/0002-view_api/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Coordinated Update to the Table API
+
+One last operation remains: the ability to turn a view into a table. This is
+necessary due to the way ArangoDB works. Unfortunately, it is not possible to
+animate a table using the results of an AQL query, thus creating a kind of
+virtual table. In turn, ArangoDB *requires* tables for constructing graphs;
+therefore, in order to use the results of a View as the basis for creating a
+network in Multinet, it is necessary to first capture the concrete results of a
+View as an actual table in the system.
```
jeffbaumes: With this information, I am failing to see the utility of views as a persistent object to be modeled in the application and database. If views cannot be used to build graphs, what would they be used for? I could see a "Create table from existing tables" functionality, where many of these operations for transformation are available, which would be built into an AQL query, executed to create a new table, then the resulting table is stored just like all other tables.

My thought would be to replace the proposed new endpoints with an augmentation to the existing endpoint:

- `POST /workspaces/<workspace>/tables`

This endpoint would now support accepting an AQL query that refers to existing tables as an alternative to a text file, and generate the table from the AQL. The AQL used to generate a table could be stored as table metadata, and possibly re-executed after data tables are updated, though editing the query after the fact seems outside MVP scope.

waxlamp: The utility I see in having a separate category for "views" is simply that they can be experimented with freely without having to generate-and-delete test tables endlessly (or, worse, pollute your workspace with tables without deleting them).

I envision a user creating a view from an AQL query, then examining the results, immediately editing the query and refining until the view generates what they are aiming for. Then, they can engage the `POST .../tables` endpoint to create the table from that view. As an advanced feature (not part of the MVP) I can imagine adding provenance to this in a simple way, capturing the work process leading to the final result.

However, adding a mode to the `POST .../tables` endpoint that directly creates a table from a query is a good idea, for cases when the user doesn't need/want to verify the results before passing them in.

This would mean a View UI, where the view objects can be manipulated and examined, as well as a couple of extra tabs in the "create table" dialog: one for creating a table from a view, and one for creating a table directly from an AQL string.

In sum, I believe the View UI will be useful to us as developers, and to end users, for experimenting with and learning AQL without having to go to the trouble of creating (possibly accidentally "big data" sized) tables.

jeffbaumes: I agree that this argument justifies having an endpoint for previewing a query before officially creating a new query-based table. But it does still not justify in my mind a wholly new model for views in the database and UI. Preview could be supported by an endpoint at the workspace level which simply executes an AQL query. This could be called while constructing the query on the client before it is finalized.

IMO a proliferation of views is at least as bad as a proliferation of tables, or perhaps even worse because the workflow would most often require a table for every view created, which is double the clutter and more for the user to keep track of. Keeping the query transient and not a top-level object would avoid additional frontend and backend complexity.

waxlamp: @AlmightyYakob, Jeff is succeeding in convincing me that an entire View API may not be justified right now. What do you think?

jjnesbitt: If there's a use case for saving queries and returning to them later for further modification, then it seems like we'd need a model to store that "View". However, if there's no use case for that, or if it's not particularly desired, then I agree that the View API is probably unnecessary. Like Jeff said, simply having the ability to create a table from a query would seem to satisfy the need for creating tables from complex ArangoDB operations. I'd think we would still want a UI component for experimenting with the AQL queries, and an ability in that UI component to turn an AQL query into a table. 

Down the road this would also facilitate the ability to create a "dynamic" or "virtual" table from the same query, which can then store information about the query anyway, among other things. [Document regarding Virtual Tables](https://docs.google.com/document/d/15WMH_FO10qDPtwu5A_uq4Ex2tWoOcVZsGmt1e-nNM3Q) (terminology might be slightly out of date).

waxlamp: Thanks for the discussion--I've updated the RFC in 5764403.

JackWilb (approved):
Looks great! I think we've settled on an excellent path forward

jeffbaumes (approved):
This proposal looks good.

waxlamp:
@AlmightyYakob, if you have any problems with this RFC, please open a pull request and start a new discussion.

PR #4, opened 2020-06-14
Add procedures for dealing with RFC rejection

waxlamp:


JackWilb (approved):

rfcs/0001-rfcs/README.md
```diff
@@ -120,6 +120,36 @@ separated by a comma. For example:
...
+
+If discussions run the other way and consensus is reached that the RFC is not
+acceptable in its current form, the author has three choices:
+
+1. Copy the `README.md` file to `my_awesome_proposal_rejected.md`, add the
+   latter file to version control, then update `README.md` in line with
+   recommendations from the discussion towards a new consensus. The title and
+   overview sections (along with details in the rest of the RFC) may need to
+   be updated, and a link to the "rejected" RFC file should be included in the
+   overview.
```
jjnesbitt, line 133: Are you saying in this case discussion is still to continue on the original PR, but we want to still keep track of old "revisions"? In that case it might be useful to have a way of numbering old revisions as well, since there could be multiple

jjnesbitt:
Also I believe an `updated` tag needs to be added to the metadata section?

rfcs/0001-rfcs/README.md
waxlamp: That's true. I can add some text about tacking on a date or something. But also at this point I don't think it's useful to worry too much about how to proceed. We already describe how other files in this repository are meant for background and context if needed; these rejected versions would just be more of that.

waxlamp:
> Also I believe an `updated` tag needs to be added to the metadata section?

I think when the initial RFC changes course (as via "rejection"), there's no need for the updated tag. It's more like the original RFC hasn't been written yet, so it's still the same author, etc.

waxlamp:
Abandoning this PR as it is superseded by RFC 0003.

Ironically, I am *not* performing a "withdrawal" via the RFC 0003 procedure, since this PR predates that process and would take further work to get it to that point.

PR #5, opened 2020-06-29
RFC: Adopt Django-esque process for creating RFCs

waxlamp:
This RFC describes a process for creating RFCs that adapts Django's methodology for creating "DEPs" (Django Enhancement Proposals). [Their process](https://github.com/django/deps/blob/master/final/0001-dep-process.rst) is detailed via its own DEP.

This PR is somewhat unusual in that it would represent a discontinuous change in our procedures. If the RFC is acceptable, I will reform the directory structure to conform to it, as well as move existing proposals to the appropriate directories (including supersession of RFC 0001).

[rendered](https://github.com/multinet-app/multinet-rfcs/tree/rfc-process/rfcs/0003-rfc-process)


rfcs/0003-rfc-process/README.md
```diff
@@ -0,0 +1,119 @@
...
+`rejection` state.
+
+**Withdrawal.** The RFC can be withdrawn by the author.
+
+1. The author determines the appropriate RFC number and renames their draft
+   directory as in the cases above.
+2. The author moves the draft directory to the `withdrawn` top level directory
+   and informs the pull request thread of their intention to withdraw.
+3. A reviewer merges the pull request.
+
```
jjnesbitt: It seems that in all of the cases here, an RFC number is always added. Is there a reason not to specify this number upon folder creation, rather than adding it in later? Is this just to reduce possible friction between simultaneous PRs?

rfcs/0003-rfc-process/README.md
```diff
@@ -0,0 +1,119 @@
...
+  kept
+
+## Proposals
+
+### New Process for RFC Creation/Review/Acceptance/Rejection
+
+1. The author clones this repository, creates a new directory in the `drafts`
+   directory with `X`s in place of the RFC number, (e.g.,
+   `drafts/XXXX-new_proposal/`), and copies the root-level `template.md` file
+   there as `README.md`.
```
jjnesbitt: If _this_ RFC is accepted, would this `template.md` be supplied after this is in the `accepted` stage?

jjnesbitt (approved):
Generally looks good, just a couple questions. Another thought, it might be good to have each directory contain it's own `README.md`, which describes what the contents of that folder represent, as well as a kind of table of contents. This would allow for easily viewing the title/descriptions of RFCs in that folder, without having to look through each individual folder to find the one they want.

rfcs/0003-rfc-process/README.md
waxlamp: Yeah I think that's exactly why. It shouldn't be a frequent problem for us but I don't think it's a big deal to do it this way either, and it will forestall trouble the first time it does happen.

rfcs/0003-rfc-process/README.md
waxlamp: Yes. I just didn't want to go to the trouble before at least preliminary buy-in.

JackWilb (approved):
LGTM

JackWilb (approved):

jjnesbitt (approved):

PR #7, opened 2020-07-08
RFC: Advanced and Robust Upload System

jjnesbitt:
Uploading a large dataset can be both timeconsuming and problematic. This RFC proposes a method for handling uploads that both facilitates ease of use, as well as reliability and fault tolerance. Uploading data is one of the most important parts of this system, and as such should be robust and featured.

[Rendered](https://github.com/multinet-app/multinet-rfcs/blob/upload-system/draft/XXXX-upload-system/README.md)

### DIFF
--- a/draft/XXXX-upload-system/README.md
+++ b/draft/XXXX-upload-system/README.md
@@ -1,0 +1,126 @@
+---
+RFC: XXXX
+Author: Jacob Nesbitt
+Status: draft
+Created: 2020-07-08
+Last-Modified: 2020-07-08
+---
+
+# RFC XXXX: Advanced and Robust Upload System
+
+Uploading a large dataset can be both timeconsuming and problematic. This RFC proposes a method for handling uploads that both facilitates ease of use, as well as reliability and fault tolerance. Uploading data is one of the most important parts of this system, and as such should be robust and featured.
+
+## Motivation
+
+Our current system of dealing with uploads is very simple and unfeatured. We send the data through a single request, with no track of progress or anything. This exposes a couple of key issues:
+
+* If the upload takes too long, the connection times out (due to a current gunicorn setting)
+* If the connection is interrupted, the upload must start over from the beginning (very frustrating on large datasets)
+* There's no progress/indication to the user. This can be implemented on the _client_ by simply tracking the HTTP request progress, but as a matter of exposing this functionality through our API, it does not exist.
+
+## Proposals
+
+### Chunked Uploads
+Instead of dealing with a single request to upload data, this data could instead be sent in chunks, allowing for many smaller HTTP requests. This would allow the following things:
+* Retain a timeout on requests, while also allowing large datasets to be uploaded with ease.
+* Fault tolerance and upload pausing/resuming
+* Progress tracking
+
+This would be accomplished through the following additions.
+
+### Hidden Tables/Networks
+To allow for uploaded chunks to be stored somewhere, when starting an upload, the desired table/network would be created in a "hidden" mode. This would prevent the table from being returned by any table listing endpoints. This table should be assumed to be volatile, as data is still being uploaded to it in this state. Whether a table is hidden or not would be tracked by a `_meta` collection on each workspace. This would contain, among other things, the information of any hidden tables in that workspace. Each document in this collection will have a structure like so:
+
+```
+{
+  collection: id,
+  hidden: bool,
+}
+```
+
+The `collection` key would point to the ID of the collection to which this document refers.
+
+Other information could be added to this collection in subsequent RFCs, making this a general purpose collection metadata storage mechanism.
+
+#### Alternative Approach
+Instead of modifying the table/network listing endpoints to omit these "hidden" tables, these endpoints could instead be modified to return objects (dictionaries), which along with the table/networks names, would include if they are temporary (i.e. being actively uploaded to). The benefit to this would be that users wouldn't be able to accidentally create duplicate tables of the same name, as they could see all tables which may exist. However, this would allow for any user (who has access to the workspace) to see an in-progress table, which may contain garbage data, and is likely unvalidated. Furthermore, if an upload fails validation after the fact, it may once again disappear from the list of tables/networks, possibly causing more confusion.
+
+### System-wide Upload Collection
+To keep track of current uploads, a `_system` collection called `uploads` would be created, which would keep track of any ongoing uploads. This collection would store, among other things, the progress of the upload, the "hidden" table(s)/graph(s) to which it belongs, and the user who initiated the upload. The schema for each document in this collection would have a structure like so:
+
+```
+{
+  current: int (bytes),
+  total: int (bytes),
+  done: bool,
+  graph: bool (indicates that a graph is being uploaded),
+  workspace: string,
+  target: {
+    tables: string[],
+    graphs: string[],
+  },
+  user: user_id (sub value),
+  _id: id (arangodb, used by clients to actually send chunks)
+}
+```
+
+### Initializing an Upload
+To initialize an upload, new endpoints will be needed, which will replace the existing upload endpoints. To initialize an upload, a request of the following format is made (endpoint for each upload format specified):
+
+```
+POST /api/upload/csv/{workspace}/{table}
+POST /api/upload/d3_json/{workspace}/{graph}
+POST /api/upload/nested_json/{workspace}/{graph}
+POST /api/upload/newick/{workspace}/{graph}
+```
+
+These endpoints replaces the original upload endpoints for each data type we currently support. This returns a JSON response, containing the upload information, in the format referenced above.
+
+A new endpoint for uploading data chunks themselves will have the following format (endpoint for each upload format specified):
+
+```
+PUT /api/upload/csv/{workspace}/{table}
+PUT /api/upload/d3_json/{workspace}/{graph}
+PUT /api/upload/nested_json/{workspace}/{graph}
+PUT /api/upload/newick/{workspace}/{graph}
+
+Body: A chunk of the original data
+
+Query Parameters: {
+  upload: id (returned by the first call),
+}
+```
+
+The `upload` query parameter is part of the information returned by the first call, and is used to denote the upload object to which the sent chunk of data belongs to. Since an upload is tied to a specific user, this endpoint is gated behind user login. Furthermore, a `PUT` request (by a logged in user) to an upload endpoint targeting an upload that does not belong to this user, should return `403 Forbidden`.
+
+
+New endpoints for getting upload information are also required. These would be of the following formats:
+
+```
+GET /api/upload/uploads
+```
+
+This endpoint is tied to the currently authenticated user, and returns all of the uploads which belong to the user. If the need for a global upload listing endpoint arises, this may have to be revised.
+
+
+### Intermediate Processing of Chunks
+
+Each data type uploader (csv, newick, etc.) will be responsible for specific logic regarding how chunked uploads are handled and stored in the "hidden" arangodb collection. Given a chunk of data, the uploader needs to be able to process/keep track the following information:
+
+* How many rows/documents are fully contained in that chunk
+* How to convert/transform a row/document from its original format, into the format required for arangodb storage
+* The current position or "context" in the full document (in graphs for example, are we currently looking at nodes?, links?, etc.)
+
+More than likely, a chunk will not often start/end exactly on the boundary between rows/documents. In this case, any "incomplete" rows/documents will need to stored in a specific location (perhaps on the upload document) until the data which completes that record arrives. Storing this data in memory is not an option, as compatibility across multiple workers must be ensured.
+
+The approach above assumes that these chunks of data are being sent **in order**. If there is a need for sending these chunks out of order, more logic would be required to keep track of "incomplete" rows/documents, and ensuring that in the future, they are able to be completed properly.
+
+#### Alternative Approach
+Another way to handle chunked uploads would be to explicitly split chunks along the boundaries of rows/documents, instead of doing this on arbitrary bytes. This would reduce the logic of chunk processing for each uploader. However, this approach would add increased complexity to, as well as place the burden of data processing on, the client application. Specific processing for each data type would be required, in order to determine the boundaries for each row/document. Furthermore, these rows/documents may be (and in practice often are) individually small, which could increase the overhead of the upload system, decreasing the overall upload speed.
+
+
+## Backwards Compatibility
+
+This RFC introduces breaking changes in the following ways:
+
+* All upload endpoints are now at a different URL, and operate in an entirely different fashion
--- a/draft/XXXX-upload-system/README.md
+++ b/draft/XXXX-upload-system/README.md
@@ -42,7 +42,7 @@ The `collection` key would point to the ID of the collection to which this docum
 
 Other information could be added to this collection in subsequent RFCs, making this a general purpose collection metadata storage mechanism.
 
-#### Alternative Approach
+#### Alternate Approach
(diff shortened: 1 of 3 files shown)


draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+```
+{
+  collection: id,
+  hidden: bool,
+}
+```
+
+The `collection` key would point to the ID of the collection to which this document refers.
+
+Other information could be added to this collection in subsequent RFCs, making this a general purpose collection metadata storage mechanism.
```
JackWilb: Just a note: This would be a good place for column metadata, too

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+Instead of modifying the table/network listing endpoints to omit these "hidden" tables, these endpoints could instead be modified to return objects (dictionaries), which along with the table/networks names, would include if they are temporary (i.e. being actively uploaded to). The benefit to this would be that users wouldn't be able to accidentally create duplicate tables of the same name, as they could see all tables which may exist. However, this would allow for any user (who has access to the workspace) to see an in-progress table, which may contain garbage data, and is likely unvalidated. Furthermore, if an upload fails validation after the fact, it may once again disappear from the list of tables/networks, possibly causing more confusion.
+
+### System-wide Upload Collection
+To keep track of current uploads, a `_system` collection called `uploads` would be created, which would keep track of any ongoing uploads. This collection would store, among other things, the progress of the upload, the "hidden" table(s)/graph(s) to which it belongs, and the user who initiated the upload. The schema for each document in this collection would have a structure like so:
```
JackWilb: When would the data here be cleaned out? When an upload fails or is successful?

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Intermediate Processing of Chunks
+
+Each data type uploader (csv, newick, etc.) will be responsible for specific logic regarding how chunked uploads are handled and stored in the "hidden" arangodb collection. Given a chunk of data, the uploader needs to be able to process/keep track the following information:
+
+* How many rows/documents are fully contained in that chunk
+* How to convert/transform a row/document from its original format, into the format required for arangodb storage
+* The current position or "context" in the full document (in graphs for example, are we currently looking at nodes?, links?, etc.)
+
+More than likely, a chunk will not often start/end exactly on the boundary between rows/documents. In this case, any "incomplete" rows/documents will need to stored in a specific location (perhaps on the upload document) until the data which completes that record arrives. Storing this data in memory is not an option, as compatibility across multiple workers must be ensured.
```
JackWilb: What is the chunking mechanism you have in mind? Is there some package you'll be leveraging or is it just the first x bytes then the second x bytes, etc.?

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+```
+POST /api/upload/csv/{workspace}/{table}
+POST /api/upload/d3_json/{workspace}/{graph}
+POST /api/upload/nested_json/{workspace}/{graph}
+POST /api/upload/newick/{workspace}/{graph}
+```
+
+These endpoints replaces the original upload endpoints for each data type we currently support. This returns a JSON response, containing the upload information, in the format referenced above.
+
+A new endpoint for uploading data chunks themselves will have the following format (endpoint for each upload format specified):
```
JackWilb: Are the data chunks going to be stored somewhere and then combined at the end or will they be uploaded as they come in?

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+### Chunked Uploads
+Instead of dealing with a single request to upload data, this data could instead be sent in chunks, allowing for many smaller HTTP requests. This would allow the following things:
+* Retain a timeout on requests, while also allowing large datasets to be uploaded with ease.
+* Fault tolerance and upload pausing/resuming
+* Progress tracking
+
+This would be accomplished through the following additions.
+
+### Hidden Tables/Networks
+To allow for uploaded chunks to be stored somewhere, when starting an upload, the desired table/network would be created in a "hidden" mode. This would prevent the table from being returned by any table listing endpoints. This table should be assumed to be volatile, as data is still being uploaded to it in this state. Whether a table is hidden or not would be tracked by a `_meta` collection on each workspace. This would contain, among other things, the information of any hidden tables in that workspace. Each document in this collection will have a structure like so:
```
waxlamp: I think we could use a term besides "hidden", as this implies something about the _intended_ visibility of the table. Really this flag is meant to explain that the table is in an intermediate state. We choose to hide such tables, but that doesn't mean the table itself is "hidden".

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+  hidden: bool,
+}
+```
+
+The `collection` key would point to the ID of the collection to which this document refers.
+
+Other information could be added to this collection in subsequent RFCs, making this a general purpose collection metadata storage mechanism.
+
+#### Alternate Approach
+Instead of modifying the table/network listing endpoints to omit these "hidden" tables, these endpoints could instead be modified to return objects (dictionaries), which along with the table/networks names, would include if they are temporary (i.e. being actively uploaded to). The benefit to this would be that users wouldn't be able to accidentally create duplicate tables of the same name, as they could see all tables which may exist. However, this would allow for any user (who has access to the workspace) to see an in-progress table, which may contain garbage data, and is likely unvalidated. Furthermore, if an upload fails validation after the fact, it may once again disappear from the list of tables/networks, possibly causing more confusion.
```
waxlamp: Yeah, I would say not to do this.

To prevent name clashes, it would be sufficient to update the table creation method to check for existing (validated and unvalidated) tables, and possibly provide a customized warning for tables that aren't visible because they're unvalidated. (`A table by that name is in process of being created` or something like that).

draft/XXXX-upload-system/README.md
waxlamp: It would be good to decide on a policy for this. We could implement it with a cron-based cleanup script that runs at midnight and deletes upload objects older than, say, 24 hours or something. (I am not sure what the right "timeout" parameters are.)

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+### Initializing an Upload
+To initialize an upload, new endpoints will be needed, which will replace the existing upload endpoints. To initialize an upload, a request of the following format is made (endpoint for each upload format specified):
+
+```
+POST /api/upload/csv/{workspace}/{table}
+POST /api/upload/d3_json/{workspace}/{graph}
+POST /api/upload/nested_json/{workspace}/{graph}
+POST /api/upload/newick/{workspace}/{graph}
```
waxlamp: Minor (but important) note--the endpoint name should be `uploads` to be consistent with the use of plural in the other REST API sections.

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+These endpoints replaces the original upload endpoints for each data type we currently support. This returns a JSON response, containing the upload information, in the format referenced above.
```
waxlamp:
```suggestion
These endpoints replace the original upload endpoints for each data type we currently support. This returns a JSON response, containing the upload information, in the format referenced above.
```

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+A new endpoint for uploading data chunks themselves will have the following format (endpoint for each upload format specified):
+
+```
+PUT /api/upload/csv/{workspace}/{table}
+PUT /api/upload/d3_json/{workspace}/{graph}
+PUT /api/upload/nested_json/{workspace}/{graph}
+PUT /api/upload/newick/{workspace}/{graph}
```
waxlamp: These should actually be `POST` endpoints, but to a different target:

```
POST /api/uploads/<upload-id>/chunk
```

Since we have upload IDs that identify the workspace/table/graph combo, there's no need to repeat them here.

This is just a sketch of how I imagine these to be. We should more or less follow Girder's practice for the design of these endpoints.

draft/XXXX-upload-system/README.md
waxlamp: Uploaded as they come in, which helps with interrupted uploads, etc., if (say) the whole DB goes down or something.

If I'm wrong here, @AlmightyYakob, please correct me.

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+The `upload` query parameter is part of the information returned by the first call, and is used to denote the upload object to which the sent chunk of data belongs to. Since an upload is tied to a specific user, this endpoint is gated behind user login. Furthermore, a `PUT` request (by a logged in user) to an upload endpoint targeting an upload that does not belong to this user, should return `403 Forbidden`.
+
+
+New endpoints for getting upload information are also required. These would be of the following formats:
+
+```
+GET /api/upload/uploads
```
waxlamp: Why not `GET /api/uploads` and `GET /api/uploads/<upload-id>`?

draft/XXXX-upload-system/README.md
waxlamp: This is the diciest part of this RFC in my opinion, and it feels like we just don't have enough understanding of traditional multipart upload schemes.

One "simple" way (air quotes because I am not sure it's this simple) is to enforce these data boundaries on the uploader logic itself. That is, the d3-json uploader would need to scan the data to deliver only completed JSON objects in each chunk. This seems pretty difficult in general, when it just feels like there ought to be a general method for collecting chunks of raw data on the server before executing the upload logic.

We will want to ask @zachmullen or other Girder wizards about this.

waxlamp: I think the problem is, the chunking has to be "semantic" in this approach. For CSVs, the chunks are easy: they are whole lines of text. For JSON formats... not sure. Perhaps JSON objects, if the JSON schema involves lists of these things? But what if they're D3-style graph objects, which have properties that are lists of nodes and edges? It's a tough problem to solve across all possible uploader types.

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+* How many rows/documents are fully contained in that chunk
+* How to convert/transform a row/document from its original format, into the format required for arangodb storage
+* The current position or "context" in the full document (in graphs for example, are we currently looking at nodes?, links?, etc.)
+
+More than likely, a chunk will not often start/end exactly on the boundary between rows/documents. In this case, any "incomplete" rows/documents will need to stored in a specific location (perhaps on the upload document) until the data which completes that record arrives. Storing this data in memory is not an option, as compatibility across multiple workers must be ensured.
+
+The approach above assumes that these chunks of data are being sent **in order**. If there is a need for sending these chunks out of order, more logic would be required to keep track of "incomplete" rows/documents, and ensuring that in the future, they are able to be completed properly.
```
waxlamp: What would be a use case that demands out-of-order chunking? I would be ok with enforcing in-order, solely to avoid this extra logic.

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+#### Alternate Approach
+Another way to handle chunked uploads would be to explicitly split chunks along the boundaries of rows/documents, instead of doing this on arbitrary bytes. This would reduce the logic of chunk processing for each uploader. However, this approach would add increased complexity to, as well as place the burden of data processing on, the client application. Specific processing for each data type would be required, in order to determine the boundaries for each row/document. Furthermore, these rows/documents may be (and in practice often are) individually small, which could increase the overhead of the upload system, decreasing the overall upload speed.
```
waxlamp: Ah this is what I was describing above, so perhaps I did not fully understand the mainline approach.

waxlamp (commented):
Looks good. We have some hairy issues to discuss here and may need to pull in other web gurus' expertise.

draft/XXXX-upload-system/README.md
zachmullen: If you want naive chunked upload, I would look to S3 as a good approximation of the state of the art. First there's an init request to create a unique ID for the upload, which validates the name (and in your case could reserve it atomically). Then chunks (S3 calls them Parts) can be uploaded in parallel and are just byte sequences that can be any size within a lower and upper bound. Once all chunks are uploaded, the client sends a completion document that lists the parts in order, and the server is responsible for "stitching" them together if you will. Basically each Part is a file with a special flag indicating it's a part, and these may be stitched together arbitrarily with a finalization request.

If you're wanting something less naive, I think the client-side chunking to split along data boundaries is a good idea. It's really not that hard to achieve.

draft/XXXX-upload-system/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+* If the upload takes too long, the connection times out (due to a current gunicorn setting)
+* If the connection is interrupted, the upload must start over from the beginning (very frustrating on large datasets)
+* There's no progress/indication to the user. This can be implemented on the _client_ by simply tracking the HTTP request progress, but as a matter of exposing this functionality through our API, it does not exist.
+
+## Proposals
+
+### Chunked Uploads
+Instead of dealing with a single request to upload data, this data could instead be sent in chunks, allowing for many smaller HTTP requests. This would allow the following things:
+* Retain a timeout on requests, while also allowing large datasets to be uploaded with ease.
```
zachmullen: There are multiple timeouts for requests to consider. Often times, for large file uploads or downloads, you want  to leave the connection open as long as it takes, and the timeout you care about is the timeout between data packets being sent. (If you're worried about security, the thing to do is to validate the content length before reading the body, and if the body goes over it, you kill the connection and abort the upload.)

draft/XXXX-upload-system/README.md
JackWilb: Maybe parallel POSTs to speed up the upload of larger files?

waxlamp: See [this comment](https://github.com/multinet-app/multinet-rfcs/pull/7#discussion_r452413501), which explains how you can do parallel chunk uploads for efficiency.

draft/XXXX-upload-system/README.md
jjnesbitt: > We could implement it with a cron-based cleanup script that runs at midnight and deletes upload objects older than, say, 24 hours or something. (I am not sure what the right "timeout" parameters are.)

I'm weary of splitting this logic between the backend and the database server (i.e. more devops burden on us). I think once an upload has completed is one easy case to mark an upload as done, and remove it. As far as dealing with incomplete lingering uploads, I'm not sure what the best approach is. I have an idea centered around using celery, which also fits into other sections of this proposal. I'm going to discuss this more offline before bringing this idea here, as it's not fully formed yet.

draft/XXXX-upload-system/README.md
jjnesbitt: :man_shrugging:, I like your idea.

jjnesbitt:
After discussion with @waxlamp we've decided to withdraw this RFC, as implementing this feature would both add a large amount of complexity, as well as take a nontrivial amount of time to complete. Since we aren't currently dealing with data at a scale which would require this level of change, we think that withdrawing this RFC until we reach that point is the best approach.

waxlamp (changes requested):
If you rename the `XXXX` to `0004`, then self-approve (is that possible?) then we can merge this as withdrawn. If you can't self-approve, we'll need @JackWilb to do so instead.

JackWilb:
I'm happy to approve when it's marked as withdrawn

jjnesbitt:
@waxlamp Self-approving is not possible, so it'll rely on you and @JackWilb to approve this.

waxlamp:
> If you rename the `XXXX` to `0004`, then self-approve (is that possible?) then we can merge this as withdrawn. If you can't self-approve, we'll need @JackWilb to do so instead.

There are a few more `XXXX`s in the files. If you `git grep` you can replace all of them.

waxlamp (approved):

JackWilb (approved):

PR #8, opened 2020-07-09
RFC: Autopublishing for MultinetJS

waxlamp:
One of the active pieces of software driving the Multinet project is
`multinetjs`, a TypeScript library providing clientside bindings and utility
functions for the Multinet API. This software grows with the needs of the
project, specifically of the [Multinet
client](https://github.com/~ltinet-client). This means that often,
a new feature in the client requires a new feature in MultinetJS, and to support
this use case, this RFC lays out a plan for more easily publishing prerelease
and development versions of the library for use in client development.

[rendered](https://github.com/multinet-app/multinet-rfcs/tree/multinetjs-autopublish/draft/XXXX-multinetjs-autopublish)


draft/XXXX-multinetjs-autopublish/README.md
```diff
@@ -0,0 +1,86 @@
...
+both pull request branches, and on master. Details for each follow.
+
+### Development Releases
+
+The idea here is to automatically publish development versions based on a PR
+branch to NPM automatically using CI. On each push from a PR branch, the GH
+Actions workflow can run the tests, perform a build, and if successful, publish
+a prerelease version to NPM. The version number can be taken from the current
+value in `package.json`, appended with the first six characters of the current
+commit's Git SHA--for example `multinet-0.47.3-a1c0ef` ([Clause
```
JackWilb: Do yarn and npm know that a version with alphanumeric appended to the end are not official releases that should be install-able unless explicitly requested? For example, in the case the 0.47.3 is the latest published version that we're supporting and 0.47.3-a1c0ef is the testing release, doing a `yarn install multinetjs` should install 0.47.3, not 0.47.3-a1c0ef

draft/XXXX-multinetjs-autopublish/README.md
```diff
@@ -0,0 +1,86 @@
...
+9](https://semver.org/#spec-item-9) of the semver spec stipulates that
+prerelease versions can be indicated by appending alphanumeric strings in this
+manner).
+
+Alternatively, instead of automatically pushing these releases, they could
+instead be done manually via an NPM script. However, this would have the
+drawback of requiring distribution of credentials to the different developers.
```
JackWilb: You mention credentials here, but not above. In the above case, with the CI doing the pushing, are the credentials stored in the GH secrets, and, thus, don't need to be distributed?

draft/XXXX-multinetjs-autopublish/README.md
waxlamp: This article is pretty old, but addresses your question (which is an excellent one): https://medium.com/@mbostock/prereleases-and-npm-e778fc5e2420.

We would need to do some experimentation to see how this works but the main idea is that a bare `yarn install multinet` will install whatever the latest published version is, but `yarn install multinet@0.47` would install `0.47.3`, not a prerelease version. Prerelease versions need to be installed explicitly, which would be the use case for Multinet client.

draft/XXXX-multinetjs-autopublish/README.md
waxlamp: Yes, and I will update the text to mention GH secrets.

waxlamp: See e5e6ff82366f617882361678e4889656c32dea1c.

subdavis:
This is an interesting idea.  I'm not sure I fully understand how it would work or what the implications are.

If I'm working on `multinet-client` that requires an un-merged feature in multinet-js and the goal is to get CI to publish client as a preview to netlify, how do I tell client about that un-merged branch's NPM prerelease?   Do I have to update package.json in client by hand after every successful prerelease of `multinet-js`?

waxlamp:
> This is an interesting idea. I'm not sure I fully understand how it would work or what the implications are.
> 
> If I'm working on `multinet-client` that requires an un-merged feature in multinet-js and the goal is to get CI to publish client as a preview to netlify, how do I tell client about that un-merged branch's NPM prerelease? Do I have to update package.json in client by hand after every successful prerelease of `multinet-js`?

The short answer is "yes".

In my view of the common case, you would do some work on multinetjs (perhaps several commits' worth) to create a new feature there, then manually change the dependency in multinet-client to match the prerelease package for the appropriate multinetjs commit. (I would likely include an NPM script in multinetjs so you could get the appropriate version string by running `yarn prerelease-version` or something like that.)

In the common case, you'd only have to do this once. If you had a strange situation where you needed to iterate on multinetjs based on what was going on in multinet-client (I can't even imagine a hypothetical case where this would happen, though I'm sure it would eventually), then you'd have to do a lot of manual updating and `yarn install`ing.

draft/XXXX-multinetjs-autopublish/README.md
JackWilb: Yeah, looks like the pre-release is tagged as latest so doing a `yarn add multinet` would give us the pre-release, but doing `yarn add multinet@0.47` would give the desired behavior. We could also consider using tags for pre-releases so they're not included in the glob when using `yarn add multinet` I'm interested to see the results of the experimentation and I'm glad it's on our radar. Thanks for the article @waxlamp

draft/XXXX-multinetjs-autopublish/README.md
```diff
@@ -0,0 +1,102 @@
...
+[semantic-release](https://github.com/semantic-release/semantic-release) can be
+used to perform automatically versioned releases of NPM projects. However, that
+approach requires a specific, machine-readable commit message format that can be
+used by the tooling to compute the correct release version.
+
+In place of that heavier-handed method, which introduces a specific burden on
+all developers, we can simply perform a release to NPM on every merge to master,
+based upon the version number listed in `package.json`. This essentially
+requires human intervention to both compute the correct semantic version for a
+given branch, and update it manually in the source. However, this also allows a
```
jjnesbitt: I'd say along with this approach, a required check in CI would be good, to ensure that the version has been bumped (failing if the version matches that of a previous release). Then, merges to master can ensure that releases go off without a hitch.

jjnesbitt (approved):
Left a comment, but LGTM

draft/XXXX-multinetjs-autopublish/README.md
```diff
@@ -0,0 +1,102 @@
...
+bit of extra control, allowing for *not* updating the version number in order to
+avoid creating a release.
+
+### Manual GitHub Actions via Workflow Dispatch
+
+GitHub Actions includes a
+[workflow_dispatch](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/)
+mechanism that allows for a pushbutton to execute a workflow. We can use this to
```
JackWilb: What does security look like for this? Can we lock this down to just admins?

JackWilb (approved):
Looks great. The workflow dispatch seems to give us the manual level of control that we need (assuming that we can specify whether it should be a pre-release or an actual release). I just left one note about security.

draft/XXXX-multinetjs-autopublish/README.md
waxlamp: I assume this is how it works, but it's something we'll have to keep in mind as we implement this.

PR #11, opened 2020-07-17
Add a "next number" file and explain how to use it

waxlamp:


JackWilb (approved):
This is probably the best solution for this. Would we be able to add a PR template that has some TODOs? I'm thinking that incrementing the counter should in there

jjnesbitt (approved):
LGTM, could also maybe add CI to ensure the number is bumped.

JackWilb:
> LGTM, could also maybe add CI to ensure the number is bumped.

This is

JackWilb:
Wrong button, sorry!

waxlamp:
> LGTM, could also maybe add CI to ensure the number is bumped.

If there's a clean and easy thing we can do, then that should be helpful. However, we're already in danger of "engineer mind" on this: if there are corner cases, or it doesn't work quite right, or whatever, then we'll end up down a rabbit hole when we could instead just follow a checklist (like Jack suggested above) and use this mechanism as a bit of support for the author.

waxlamp:
> This is probably the best solution for this. Would we be able to add a PR template that has some TODOs? I'm thinking that incrementing the counter should in there

I like this idea, but I also don't want to pour much time at all into developing one, especially while we're still kind of shaking out problems with our approach.

PR #12, opened 2020-07-28
Multinet ORM

jjnesbitt:
This RFC proposes an ORM for dealing with database operations in Multinet.


draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+##### User
+```
+# Previously used the term "cookie"
+def get_session() -> str: ...
+def set_session(): ...
+def delete_session(): ...
+
+# Serializes the object to JSON
+def serialize() -> Dict: ...
```
waxlamp: Presumably, all the ORM classes should have a `serialize()` method, right? That would let us construct different types of JSON responses to various levels of API endpoints, in a more comprehensive way than we do now.

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+def register(user: UserInfo) -> User: ...
+def from_session(session_id: str) -> User: ...
+```
+
+##### Table
+```
+def rows() -> Iterable[Dict]: ...
+def row_count() -> int: ...
+def keys() -> Iterable[str]: ...
+def edge() -> bool: ...
```
waxlamp:
```suggestion
def is_edge() -> bool: ...
```

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Static
+def exists(user: UserInfo) -> bool: ...
+def register(user: UserInfo) -> User: ...
+def from_session(session_id: str) -> User: ...
+```
+
+##### Table
+```
+def rows() -> Iterable[Dict]: ...
```
waxlamp: This method should take limit and offset parameters.

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+# Previously used the term "cookie"
+def get_session() -> str: ...
+def set_session(): ...
+def delete_session(): ...
+
+# Serializes the object to JSON
+def serialize() -> Dict: ...
+
+# Static
+def exists(user: UserInfo) -> bool: ...
```
waxlamp: Does this method test whether a user by the supplied `sub` value exists? How do you use an entire `UserInfo` to ask about whether that user exists?

I think we might also want a static method that performs a search by email address. What do you think?

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+def register(user: UserInfo) -> User: ...
+def from_session(session_id: str) -> User: ...
```
waxlamp: What does this method do?

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+##### Graph
+```
+def nodes() -> Iterable[Dict]: ...
+def node_tables() -> Iterable[Table]: ...
+def edge_table() -> Table: ...
+```
+
+
+##### Workspace
+The workspace class would contain the following methods:
```
waxlamp:
```suggestion

```

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+def create_table(name: str, rows: Iterable[Dict], edge: bool = False) -> Table: ...
+def create_aql_table(name: str, aql: str) -> Table: ...
+def delete_table(name: str): ...
+
+def create_graph(
+  name: str,
+  edge_table: str,
+  from_collections: Iterable[str],
+  to_collections: Iterable[str],
```
waxlamp: Are these necessary? We can automatically figure out what values these should have, so are we simply doing that somewhere else, and leaving this method to match up with the Arango-level functions for creating a graph?

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+) -> Graph: ...
+def delete_graph(name: str): ...
+
+def permissions() -> Dict: ...
```
waxlamp: We need a way to *set* permissions as well.

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+# Static
+def create(name: str) -> Workspace: ...
+def exists(name: str) -> bool: ...
+```
+
+
+### Open Questions
+
+* Should nodes be included in the ORM, or should they remain as dictionaries?
```
waxlamp: I think we can probably leave them as dicts, at least for now. Nodes (and rows) are "raw" objects in that there aren't really mutable operations to perform on them--they're just primitive data.

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+* Should `create`, `delete`, etc. be attached to the classes they are operating on? This was the original proposal, but after discussion was decided against. The reason it's mentioned here is to highlight the benefit of such a model, mainly:
```
waxlamp: I would say no. Arango forces the database object to create its collections, so we should mirror that in our ORM. To attach these as static methods of e.g. `Table` would mean supplying the workspace object to those methods, which just seems weird and confusing to me.

There's probably an argument that the concept of ORM means we *should* put these methods on the classes themselves, but I think it's important to model how we will actually conceptualize and call these methods in our codebase, where our own API already implies that the, e.g., `Table`-level operations cannot be carried out without first being in a `Workspace` context.

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+  * Methods are more tightly coupled to the classes they are affecting
+  * Methods are spread more evenly between classes, instead of `Workspace` containing the bulk of the operations
```
waxlamp: I don't consider this to be a problem in and of itself.

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+def create(name: str) -> Workspace: ...
+def exists(name: str) -> bool: ...
+```
+
+
+### Open Questions
+
+* Should nodes be included in the ORM, or should they remain as dictionaries?
+* Should `create`, `delete`, etc. be attached to the classes they are operating on? This was the original proposal, but after discussion was decided against. The reason it's mentioned here is to highlight the benefit of such a model, mainly:
+  * Methods are more tightly coupled to the classes they are affecting
```
waxlamp: As I mentioned above, you still need to say which workspace is doing the creating, which means that these methods will actually *not* be as tightly coupled to the class as this method membership would imply.

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+  * Methods are spread more evenly between classes, instead of `Workspace` containing the bulk of the operations
+
+  However, the main reason against this approach is that it involves instantiation of objects possibly before they exist, which may be confusing.
```
waxlamp: This is more evidence that the methods should stay with `Workspace`. Just to drive the main point home: we are modeling the manipulation of Arango collections from Arango databases; in that context, there's no notion of creating a collection without first being in the context of a workspace. So to shift over to Multinet's idea of "workspaces" and "tables", to introduce this idea of tables preceding their workspaces is indeed confusing, and just isn't needed in order to make this work.

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Backwards Compatibility
+
+Ideally, this change has no effect on the API, as it is focused on changing the way our database operations are structured, not their behavior.
```
waxlamp: You can make this statement stronger (by cutting out the qualifiers): it will be a design goal not to change the API because of this code change.

waxlamp (changes requested):
Looks good. I left several suggestions for how to change the proposed ORM API. And if the questions you listed are resolved, then I would say to strike that whole section (possibly updating other parts of the text if the discussion is valuable enough to actually record in some way).

draft/XXXX-orm/README.md
jjnesbitt: An example of where you'd use the entire `UserInfo` is when the user initially logs in through OAuth. We'll have the user information, but they wont exist in our database yet. If we were to have the sub value, we'd probably have the rest of the information as well.

I think a `search` method is a good idea, I can add it.

draft/XXXX-orm/README.md
jjnesbitt: Returns a user from the specified session id (stored in the user's browser via cookies).

draft/XXXX-orm/README.md
jjnesbitt: Good point.

draft/XXXX-orm/README.md
jjnesbitt: This is a good point. I was already convinced on your methodology for distributing methods, but this finally made it clear as day to me. That is, in my alternative approach, you would be able to, for example, run `Table(name).create()` without ever specifying a workspace, which is invalid.

draft/XXXX-orm/README.md
jjnesbitt: I stubbed this out based on the [existing function](https://github.com/multinet-app/multinet-server/blob/264089818733eb3bbcad00495a9526f7ccc487ea/multinet/db.py#L474-L500):

But you're right, when we're calling that function, we're just passing in information that it could derive anyway. I'll update the definition.

draft/XXXX-orm/README.md
jjnesbitt: I suppose, but there's some open questions regarding that. Like, does the `serialize` method dump everything into JSON, or is it selectively outputting information? And where do we plan on using it?

draft/XXXX-orm/README.md
waxlamp: Oh yes that makes total sense. The full calling sequence `User.from_session(session_id)` makes it very clear.

draft/XXXX-orm/README.md
waxlamp: Well, maybe there's good reason to retain this stuff at the DB layer? If we compute it automatically in the API layer or something, we'd still need to pass the info along to ArangoDB.

jjnesbitt: I'm not sure what you mean. Currently, in the [D3 JSON uploader](https://github.com/multinet-app/multinet-server/blob/264089818733eb3bbcad00495a9526f7ccc487ea/multinet/uploaders/d3_json.py#L124-L132), we're calling:

```
properties = util.get_edge_table_properties(workspace, edge_table_name)
```

and then immediately passing the result into `db.create_graph`. We also pass `workspace` and `edge_table_name` into `db.create_graph`, so there's no need to make this call happen outside of the `create_graph` function. That's all I was saying.

draft/XXXX-orm/README.md
jjnesbitt: Regarding search: 98ab5ff

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,104 @@
...
+##### Workspace
+```
+
+def table(name: str) -> Table: ...
+def tables() -> Iterable[Table]: ...
+
+def graph(name: str) -> Graph: ...
+def graphs() -> Iterable[Graph]: ...
+
+def metadata() -> Dict: ...
```
JackWilb: Does this return all the metadata for the workspace? It might be worth considering adding a method for table metadata that we will fill once we have a more mature uploader

jjnesbitt: I agree, but since we don't really know what that looks like yet, I think we should wait and update this RFC when the time comes.

JackWilb: Agreed

draft/XXXX-orm/README.md
waxlamp: Ok. Let's leave it out for now.

BTW, your comment says "to JSON" but the return type is `Dict`. Both of those can't be correct.

draft/XXXX-orm/README.md
waxlamp: What I'm saying is, to keep in pace with how we already do things, we *do* need `from_collections` and `to_collections` in this method (answering my initial question with a "yes, these are necessary).

In other words, I think this is fine as is, and there's no need to "update the definition", as you put it. If that's correct, we can resolve this conversation without further ado.

draft/XXXX-orm/README.md
jjnesbitt: Good point. I'd think serializing directly to JSON (return type of `str`) would be the best approach, as it would remove dealing with dictionaries and handling the associated typing problems.

waxlamp: I'd be fine with that. If you just change the return type to `str` then we can resolve this convo.

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,104 @@
...
+```
+def rows(offset: int = 0, limit: int = 0) -> Iterable[Dict]: ...
+def row_count() -> int: ...
+def keys() -> Iterable[str]: ...
+def is_edge() -> bool: ...
+```
+
+##### Graph
+```
+def nodes() -> Iterable[Dict]: ...
```
waxlamp: I think this needs `offset` and `limit` parameters too, right?

draft/XXXX-orm/README.md
```diff
@@ -0,0 +1,104 @@
...
+def node_tables() -> Iterable[Table]: ...
+def edge_table() -> Table: ...
+```
+
+
+##### Workspace
+```
+
```
waxlamp:
```suggestion
```

waxlamp (changes requested):
I think this is just about ready. I'd strike out the whole "open questions" section, address my comment about the tone of the backwards compatibility section, and I just have the one remaining suggestion about a return type.

draft/XXXX-orm/README.md
jjnesbitt: 24fe4e1

draft/XXXX-orm/README.md
jjnesbitt: 9cb6ad8

draft/XXXX-orm/README.md
jjnesbitt: c644f93

waxlamp (approved):

JackWilb (approved):

PR #13, opened 2020-08-04
RFC: Add multipart upload

waxlamp:


waxlamp:
@mvandenburgh, take a look at this and let me know if anything is unclear. Feel free to contact me on Slack to discuss in detail.

draft/XXXX-multipart-upload/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Summary of Approach
+
+The basic idea is as follows: the client initiates a new upload session for a
+given file; this action causes the server to create a new collection in the
+database. Then the client issues one or more "chunk" uploads, each sending a
+range of bytes, up to some size limit, to the server. The server saves each
+chunk as a document in the database collection associated with this file upload.
+Finally, the client issues a "finalize" call, which marks the database
+collection as being complete.
```
JackWilb: It's generally not advisable to store large files in the db, but instead to store them on disk and to just store a filepath in the db. Is this an option for us? What is the benefit of storing the files in the db? It seems like this increases the number of db transactions a lot

draft/XXXX-multipart-upload/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+The server can then ask for a file-like object associated with the upload. The
+server can then use this object to perform validation, retrieve table rows, and
+settle them in their final home in the server (as is done currently, directly
+with the file contents).
```
JackWilb: Don't we need the whole file for validation? How do you plan to get the whole file out of the db from the multiple parts?

draft/XXXX-multipart-upload/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+### API design
+
+This RFC adds an `uploads` section to the Multinet API, which contains the
+following endpoints:
+
+- `POST /api/uploads` - creates a new upload object by selecting a random,
+non-preexisting SHA string, then creating an Arango collection keyed by the
+string. This endpoint returns the ID string.
+- `POST /api/uploads/<upload-id>/chunk?sequence=<sequence-number>` - uploads one
```
JackWilb: It's uncommon to use a querystring for a POST. It would be more standard to have `POST /api/uploads/<upload-id>/chunk/sequence/<sequence-number>`

draft/XXXX-multipart-upload/README.md
```diff
@@ -0,0 +1,115 @@
...
+chunk of data to the named upload object. This endpoint returns the sequence
+number, and may raise various status codes to indicate problems.
+- `POST /api/uploads/<upload-id>/finalize` - finalizes the IDed upload by
+preventing further calls to `POST /api/uploads/<upload-id>/chunk` and performing
+a checksum on the settled data.
```
JackWilb: What happens if a chunk upload fails half way? Would there be a hanging part? How would the API stitch together the file with these half uploaded chunks?

Also, what is the checksum here? A checksum for each chunk? Maybe that should be ran on each post of a chunk

draft/XXXX-multipart-upload/README.md
```diff
@@ -0,0 +1,115 @@
...
+to remove the intermediate upload data, leaving the server in a clean state.
+
+## Questions
+
+1. How do we detect upload interruptions? What is the right way to resume such
+   an interrupted upload?
+2. What do we do with "stale" upload objects? One answer is to have a `vaccum`
+   endpoint that looks for upload objects older than a specific cutoff age and
+   remove them, then run that endpoint periodically. Better still, it could be
+   installed as a cronjob on the server directly.
```
JackWilb: A cronjob seems appropriate

JackWilb (commented):
Some preliminary thoughts

draft/XXXX-multipart-upload/README.md
waxlamp: We might be able to do it by storing files on the AWS instance in some structured way, but I'm not sure what interface we'd use to manage file transfers from Python. The advice to use the DB came from others at Kitware who have done this sort of thing before.

I'm not really worried about the number of DB transactions, because the vast majority of the time the upload process will use a single chunk, and in any case these extra DB transactions because of upload will (1) be linear in the size of the files being uploaded and (2) only occur a constant number of times.

Are there other reasons it's inadvisable to store files in the DB?

draft/XXXX-multipart-upload/README.md
waxlamp: The idea is that the `UploadReader` class will give you a file-like interface to the chunks. So you start out at position 0 in the "file", which translates to something like "chunk 0, byte 0". Then a `.read()` call would grab as many bytes as needed, possibly loading more chunks as the virtual file position advances past chunk boundaries.

If that class is available, then you can simply call, e.g., `.readline()` on it to extract lines of a CSV file, or `.read()` followed by `json.loads()` to parse it into JSON, or make use of a streaming JSON reader library, or whatever else we're already doing on the request body currently.

draft/XXXX-multipart-upload/README.md
waxlamp: Do you have a reference for this? As a counterexample, the Girder API has plenty of `POST` endpoints that accept query arguments.

The alternative you gave has different RESTful semantics on the face of it: it's saying "create a `sequence` object with ID `<sequence-number>`.

draft/XXXX-multipart-upload/README.md
JackWilb: RAM usage is one thing that springs to mind, but we can cross that bridge when we get there. I do think accessing the files from outside the database would be more tricky. You'd have to have a webserver that can serve static data from a directory, which is a ton more overhead. Storing in the database seems like a good solution for now.

draft/XXXX-multipart-upload/README.md
waxlamp: Chunks are all-or-nothing, so if a chunk upload fails halfway, then nothing is written to the database, and the client receives an error status code.

There are block-based checksums that can be run by streaming the target data. My thinking was to compute such a checksum and compare it to one sent at some point during this whole process.

Your idea for running checksums on each chunk is not a bad one. However, to make use of it, the client would need to compute a checksum for each chunk, and send that value along with the chunk upload. That's a very unnatural way to operate. Usually, the client is interested in verifying the checksum for the full data. That can certainly be computed once and sent along with the `finalize` call.

draft/XXXX-multipart-upload/README.md
waxlamp: We'll want to figure out if there's an easy way to incorporate that into provisioning. I believe it is pretty simple--in particular, that Ansible includes modules for doing this. And then we should discuss whether cronjobs are the best way to handle this.

draft/XXXX-multipart-upload/README.md
JackWilb: This was just my experience. It does seem that it's allowed to do a POST with a query string, just uncommon. See [here](https://stackoverflow.com/questions/611906/http-post-with-url-query-parameters-good-idea-or-not).

Your point about about the semantics is correct. Is that not what this endpoint is doing? Are all the sequence numbers determined ahead of time?

draft/XXXX-multipart-upload/README.md
jjnesbitt: If we're dealing with larger files often, I think delegating this to something like AWS S3 (which has good support for multipart upload) is a good idea, since it seems like a more robust and efficient way to upload the data. It would also let us limit the amount of ram we're using when processing the file, while keeping a simple interface for reading the file. I.e., create upload, return redirect to S3, which does upload in multiple parts, and then when the finalize request is sent to our server, we reach out to the file, and stream the bytes in to process the now fully constructed file.

However, all of that being said, I don't think we need to deal with that right now, as our files are small enough for the time being. I mainly wanted to convey the idea, to consider in case we come to a point where this implementation doesn't scale well. I'd think as the file size increases, the difference in upload performance would start to be noticeable. Plus, some of this stuff would need to be laid out for the S3 path anyway, so if we do decide to re-examine that approach in the future, some of the ground work will already be done.

draft/XXXX-multipart-upload/README.md
jjnesbitt: While I think on smaller files, you could read in the whole file for validation no problem, ideally we'd want to change the way we validate files to stream the data from any data source, including an `UploadReader`. This is mentioned in https://github.com/multinet-app/multinet-server/issues/408.

draft/XXXX-multipart-upload/README.md
```diff
@@ -0,0 +1,115 @@
...
+instantiate an `UploadReader` object. This file-like object can be used to
+perform validation and bulk ingestion into a Multinet data table proper. The
+main advantage of this usage is that an `UploadReader` does not exhaust its data
+iterator after one pass; it can be rewound by using the `.seek()` method. This
+means that both validation and ingestion can be performed without a memory
+blowup caused by exhausting an HTTP request body iterator into a string in
+memory.
+
+Finally, the client (or the uploader methods) can invoke the `DELETE` endpoint
+to remove the intermediate upload data, leaving the server in a clean state.
```
jjnesbitt: It seems a bit odd to me to put the client in charge of deleting the old upload data, after it's been validated and re-stored, as it's more of an implementation detail. That is, all the client cares about is that it's finished uploading its file, at which point it calls the `finalize` endpoint, and expects that after that request finishes, the data it cares about is validated and stored. So, I think deleting the old upload data should just be the final step of the `finalize` process.

This raises another question, which is that currently, we're depending on the fact that the processing/validation of uploads is relatively quick (i.e. I don't think we expect calls to the `finalize` endpoint to be taking too long). If we start processing larger data, this may not be the case, at which point something like background processing of a dataset may be necessary.

Again, I don't think we need to do this now, just something to keep on the radar.

draft/XXXX-multipart-upload/README.md
```diff
@@ -0,0 +1,115 @@
...
+
+## Questions
+
+1. How do we detect upload interruptions? What is the right way to resume such
+   an interrupted upload?
```
jjnesbitt: I don't really think we need to "detect" upload interruptions (from the server, that is), we just need to have a way to resume the uploads. This would necessitate another couple endpoints:

* `GET /api/uploads` - List all active uploads for the current user
* `GET /api/uploads/<upload-id>` - Get the metadata about a specific upload

So that, if the upload is interrupted or something, the client can get this list of active uploads, and show the user, so that they may manually resume this upload. The second endpoint mentioned above may not be strictly necessary for this, but may be a good idea to have.

draft/XXXX-multipart-upload/README.md
```diff
@@ -0,0 +1,115 @@
...
+   strategy allows for resuming such uploads without losing the time already
+   spent in uploading one more initial chunks of the file.
+
+## Proposals
+
+### Summary of Approach
+
+The basic idea is as follows: the client initiates a new upload session for a
+given file; this action causes the server to create a new collection in the
+database. Then the client issues one or more "chunk" uploads, each sending a
```
jjnesbitt: Which database would these new per-upload collections be stored? The `_system` database? Or maybe we want a separate `_upload` database? I also think we'll need a separate collection for tracking the status of each upload (one-to-one with the collections in whatever database upload collections are stored). Each document in this collection would track the metadata, status, etc. of each upload, and would point to a collection in the "uploads" database.

draft/XXXX-multipart-upload/README.md
JackWilb: I think checking the whole file is okay, I just wonder if it will be easy to compute for the whole file if it's broken up into chunks

jjnesbitt (changes requested):
Generally on-board, just have comments/concerns about specifics.

draft/XXXX-multipart-upload/README.md
waxlamp: I agree with keeping S3 in our back pocket.

However, I don't understand why memory usage is a problem. The whole point of multipart upload is to never have more than a certain sized chunk of data in flight at once. Just for illustration with made up numbers: rather than holding a 1GB file in memory for validation or uploading or whatever, we'd perform multiple operations on 1MB chunks 1000 times. Settling 1000 1MB chunks into ArangoDB would be a lot of DB connections, but it seems completely limit the memory requirements for the operation.

draft/XXXX-multipart-upload/README.md
jjnesbitt: I think ideally we'd have a hands-on system in place for dealing with stale uploads, instead of using a cron job, as it introduces a devops burden, as well as a disconnect between our application and the database (i.e. the database is modifying things behind the server's back). Plus, it makes the application less portable, as if you were using some managed database for example, you wouldn't be able to just throw a cron job on there to deal with this. However, I'm not sure what we can really do at this point besides that, without adding more complexity anyway (like celery, etc.).

I've done cron job provisioning through Ansible before, and it isn't difficult, but does introduce some very deployment-specific devops details.

draft/XXXX-multipart-upload/README.md
waxlamp: > This was just my experience. It does seem that it's allowed to do a POST with a query string, just uncommon. See [here](https://stackoverflow.com/questions/611906/http-post-with-url-query-parameters-good-idea-or-not).
> 

One thing that might have gotten lost is that we *are* going to send the chunk data in the body of the request. The SO answer seems concerned with the oddness of not doing that (though, I'm not really sure why that should be so odd). Query arguments seem to be the natural way for us to express the extra data in this case. And I need to update the RFC description with the bit about the request body.

> Your point about about the semantics is correct. Is that not what this endpoint is doing? Are all the sequence numbers determined ahead of time?

Usually in a REST API, the "path elements" in an endpoint refer to actual resources. In this case, we've created an "upload" resource, and with the `/chunk` endpoint we're modifying that object by adding each data chunk. The chunks need to have sequence numbers, but those numbers are like chunk metadata, not resources in and of themselves. That's why I think sending that in a query argument (rather than path elements) is perfectly natural.

draft/XXXX-multipart-upload/README.md
jjnesbitt: I don't think there's really memory concerns regarding this part of the upload process, just performance concerns really. If we have to chunk a file into 1000 parts, that's a lot of overhead, and thus the upload will be slow. Plus, arango isn't really designed to store binary data. However I don't really know if performance will be an issue in practice, it's just something in my mind.

draft/XXXX-multipart-upload/README.md
waxlamp: > I think checking the whole file is okay, I just wonder if it will be easy to compute for the whole file if it's broken up into chunks

For block-based checksums it should be easy (though I don't have experience with this). If there's a Python library that does it, I imagine simply passing an iterator to a function would "just work". This is something we need to do a bit of research into.

draft/XXXX-multipart-upload/README.md
JackWilb: >One thing that might have gotten lost is that we are going to send the chunk data in the body of the request. 

I understood that. I was just thinking in terms of multi-part forms where all data gets sent in the POST body, not split into the query string and the body

You've convinced me that this makes sense and that the syntax of the requests is appropriate

jjnesbitt: In my experience, the use of query parameters is independent of the HTTP Method used.

draft/XXXX-multipart-upload/README.md
waxlamp: > It seems a bit odd to me to put the client in charge of deleting the old upload data, after it's been validated and re-stored, as it's more of an implementation detail. That is, all the client cares about is that it's finished uploading its file, at which point it calls the `finalize` endpoint, and expects that after that request finishes, the data it cares about is validated and stored. So, I think deleting the old upload data should just be the final step of the `finalize` process.
> 

The problem is, the `finalize` call can't necessarily delete the data because we still need the "uploader" to do its thing. It would be possible to tell the `finalize` call the name of the uploader we want to run (which could then run the uploader and on success, perform the deletion), but that seems like poor software design, since the uploaders themselves might need to be passed data or be called in arbitrary ways. We'll have to figure out how to do this properly. (Maybe the uploader itself can perform the deletion as a final step.)

> This raises another question, which is that currently, we're depending on the fact that the processing/validation of uploads is relatively quick (i.e. I don't think we expect calls to the `finalize` endpoint to be taking too long). If we start processing larger data, this may not be the case, at which point something like background processing of a dataset may be necessary.
> 
> Again, I don't think we need to do this now, just something to keep on the radar.

Good point. You can file an issue if you want to, though this problem will likely be obvious when it occurs, and at that time we can file issues, discuss what to do, etc.

draft/XXXX-multipart-upload/README.md
waxlamp: This feels right.

So when this happens, and the client shows the user a list of uploads, what does the user do? They select an upload, and then supply the file to upload once more?

draft/XXXX-multipart-upload/README.md
waxlamp: I think an `_uploads` database is a good idea. Did we already figure out that collections can't contain metadata in ArangoDB? We might wind up with an `upload_mapping` collection in `_system` in that case (I think this is what you're describing).

draft/XXXX-multipart-upload/README.md
waxlamp: > Are all the sequence numbers determined ahead of time?

Forgot to answer this above.

I believe the client determines these by matching up slices of the file it's sending with their ordering, so typically you just go `0` for the first chunk, `1` for the second, etc.

draft/XXXX-multipart-upload/README.md
waxlamp: I actually agree with your concerns. However, we're already chest-deep in devops-specificity, and we need some way to solve this problem. I'm totally open to ideas.

When you say "hands on", do you mean that we should execute a script manually every so often to clean up after stale uploads? If we have such a thing, why wouldn't we just automate that via a devops-deployed cronjob?

draft/XXXX-multipart-upload/README.md
waxlamp: In my response I forgot to mention: I agree that it's odd to have the client issue the request for deletion. It ought to be done automatically the moment the intermediate data is no longer needed (specifically, without any extra input from the client).

draft/XXXX-multipart-upload/README.md
jjnesbitt: Sorry, by "hands on", I meant that the server application is aware/in control of the methods used to clean up stale uploads.

draft/XXXX-multipart-upload/README.md
jjnesbitt: I don't believe that you can set arbitrary data on collections. What you're describing in the end of your comment is what I had in mind, although I'd prefer to get away from the semantics of "mapping" and just call it "uploads" or something, which also happens to reference a collection.

draft/XXXX-multipart-upload/README.md
jjnesbitt: Yeah, that sounds right. I'm assuming the data-model for the "status" of an upload includes a byte-count or something similar, so when the user selects a file to resume uploading, the client just starts from that byte-count and resumes sending chunks.

I suppose as part of this, we should properly flesh out what data that status contains. I can think of a few fields off the top of my head:

```
{
  "id": <uuid>,
  "uploaded": <int>,       # represents bytes uploaded
  "collection": <string>,  # collection that stores the upload data
  ...
}
```

I think this would be stored in the `_system` "uploads" collection that I mentioned elsewhere.

draft/XXXX-multipart-upload/README.md
```diff
@@ -0,0 +1,115 @@
...
+data in the database in an intermediate form.
+
+Next, the client can invoke one of the uploader paths, which will in turn
+instantiate an `UploadReader` object. This file-like object can be used to
+perform validation and bulk ingestion into a Multinet data table proper. The
+main advantage of this usage is that an `UploadReader` does not exhaust its data
+iterator after one pass; it can be rewound by using the `.seek()` method. This
+means that both validation and ingestion can be performed without a memory
+blowup caused by exhausting an HTTP request body iterator into a string in
+memory.
```
mvandenburgh: This would address https://github.com/multinet-app/multinet-server/issues/408 and prevent the need to refactor those validation functions. Those validation functions are used elsewhere IIRC, so not having to modify those/create new ones would be great to minimize code complexity

draft/XXXX-multipart-upload/README.md
jjnesbitt: > The problem is, the `finalize` call can't necessarily delete the data because we still need the "uploader" to do its thing. It would be possible to tell the `finalize` call the name of the uploader we want to run (which could then run the uploader and on success, perform the deletion), but that seems like poor software design, since the uploaders themselves might need to be passed data or be called in arbitrary ways. We'll have to figure out how to do this properly. (Maybe the uploader itself can perform the deletion as a final step.)

I don't understand why the finalize call can't delete the data after uploads completes. To me, the finalize call has roughly the following steps:

1. Call the appropriate uploader by passing an instance of `UploadReader` (along with any necessary metadata) to the appropriate filetype uploader (the upload filetype could be either determined at finalization, or supplied by the client when initially creating the upload. I'm in favor of the latter.)

2. Once the previously called uploader completes, it has either validated and inserted that upload into the appropriate place, or has failed and thrown a validation error, which is just passed up the call stack and returned as the response to the client. If it's the former case, we've successfully inserted the uploaded data into its target location, without error, and can now delete the upload.

draft/XXXX-multipart-upload/README.md
mvandenburgh: I agree on using Arango for the time being and keeping S3 in mind potentially for the future if there's scalability issues (especially if we're not dealing with large files often). I'm not too familiar with ArangoDB but I wonder what write speeds will look like for the binary data in practice.

draft/XXXX-multipart-upload/README.md
mvandenburgh: > For block-based checksums it should be easy (though I don't have experience with this). If there's a Python library that does it, I imagine simply passing an iterator to a function would "just work". This is something we need to do a bit of research into.

Just from a quick search, it seems pretty simple, I don't think it'll be a problem.
https://stackoverflow.com/a/3431838

draft/XXXX-multipart-upload/README.md
mvandenburgh: Is the database hosted on heroku? If so, maybe this could be another option next to Ansible
https://devcenter.heroku.com/articles/scheduler

That does make things dependent on Heroku as a host though so maybe it's not worth it. Still, thought it was at least worth mentioning.

draft/XXXX-multipart-upload/README.md
mvandenburgh: I agree that having the client be responsible for deletion isn't ideal. I like @waxlamp's idea to have the uploader perform the deletion as the final step, although full disclosure I'm not 100% familiar with the uploaders yet, so I might be overlooking some issues with that approach.

mvandenburgh (commented):
I left some comments and replies below, but in general I think multipart upload is the best way to deal with the issues that come with handling large file uploads.

draft/XXXX-multipart-upload/README.md
waxlamp: Ok. I'll add some text about this.

I think the model for this will be a bit more complex than simply what prefix of bytes has been uploaded. One benefit of multipart upload is that you can send several chunks in parallel. So if an "earlier" chunk fails while a "later" one succeeds, we'd need some way to encode that.

draft/XXXX-multipart-upload/README.md
waxlamp: Yeah I agree. You could even file an issue to refactor `workspace_mapping` to `workspaces` or something (though that would be pretty low priority).

waxlamp:
@mvandenburgh wrote:
> Is the database hosted on heroku? If so, maybe this could be another option next to Ansible
https://devcenter.heroku.com/articles/scheduler

Unfortunately, no. We're hosting that on AWS. Heroku doesn't have an offering for ArangoDB, I assume because the ArangoDB people themselves sell hosted services. Unfortunately (again), the ArangoDB hosting services offer a minimum service level that's much too large/expensive for our purposes.

draft/XXXX-multipart-upload/README.md
waxlamp: > > The problem is, the `finalize` call can't necessarily delete the data because we still need the "uploader" to do its thing. It would be possible to tell the `finalize` call the name of the uploader we want to run (which could then run the uploader and on success, perform the deletion), but that seems like poor software design, since the uploaders themselves might need to be passed data or be called in arbitrary ways. We'll have to figure out how to do this properly. (Maybe the uploader itself can perform the deletion as a final step.)
> 
> I don't understand why the finalize call can't delete the data after uploads completes. To me, the finalize call has roughly the following steps:
> 
> 1. Call the appropriate uploader by passing an instance of `UploadReader` (along with any necessary metadata) to the appropriate filetype uploader (the upload filetype could be either determined at finalization, or supplied by the client when initially creating the upload. I'm in favor of the latter.)

I also agree with using the latter.

> 2. Once the previously called uploader completes, it has either validated and inserted that upload into the appropriate place, or has failed and thrown a validation error, which is just passed up the call stack and returned as the response to the client. If it's the former case, we've successfully inserted the uploaded data into its target location, without error, and can now delete the upload.

Yeah, I like this idea.

Would the finalize call _require_ a filetype? That would mirror current practice where by necessity you have to select a file type in order to upload a file.

What do we do if the upload fails to pass validation? We have no way to alter the data on the server, so it seems like the only thing we can offer currently is to show the user the errors, and let them fix their data and try again. Once again, this mirrors current practice so I think it would be a fine place to start. In that case, we could also delete the upload on failure.

It would be a very advanced (and very cool) feature to allow in-place editing of uploaded data, but it's out of scope I think. We'd be entering shark-infested waters, since it's an open research/industry question of how to write a viable "data wrangler" tool.

draft/XXXX-multipart-upload/README.md
waxlamp: > Is the database hosted on heroku? If so, maybe this could be another option next to Ansible
> https://devcenter.heroku.com/articles/scheduler
> 
> That does make things dependent on Heroku as a host though so maybe it's not worth it. Still, thought it was at least worth mentioning.

Unfortunately, no. We're hosting that on AWS. Heroku doesn't have an offering for ArangoDB, I assume because the ArangoDB people themselves sell hosted services. Unfortunately (again), the ArangoDB hosting services offer a minimum service level that's much too large/expensive for our purposes.

waxlamp: > Sorry, by "hands on", I meant that the server application is aware/in control of the methods used to clean up stale uploads.

How do we accomplish this? This sounds like an alternative plan to using cronjob via devops deployment of the DB.

draft/XXXX-multipart-upload/README.md
jjnesbitt: @waxlamp If we've reached a consensus here (which I think we have), can you add the points from this discussion into the RFC?

waxlamp: Yes, I've been meaning to complete the work on this RFC. I'll try to focus on it today.

I'll also add the PR Mike's working on to the "reference implementation" section.

draft/XXXX-multipart-upload/README.md
jjnesbitt: > Would the finalize call _require_ a filetype? That would mirror current practice where by necessity you have to select a file type in order to upload a file.

I thought the filetype would be specified on upload creation, is that correct?

> What do we do if the upload fails to pass validation? We have no way to alter the data on the server, so it seems like the only thing we can offer currently is to show the user the errors, and let them fix their data and try again. Once again, this mirrors current practice so I think it would be a fine place to start. In that case, we could also delete the upload on failure.

Yeah I think this is where we should start, as to not introduce too much complexity at once. However, I think it's important that we properly inform the user of the errors with an upload, so that it doesn't just fail for no reason.

> It would be a very advanced (and very cool) feature to allow in-place editing of uploaded data, but it's out of scope I think. We'd be entering shark-infested waters, since it's an open research/industry question of how to write a viable "data wrangler" tool.

I agree this would be a neat feature, and would require at least a few things:

* A data editor on the client to edit the uploaded data (and would differ by file-type)
* A paginated endpoint that returns the data from an upload
* An interface between the db and api layer to stream upload data (perhaps using the `UploadReader` class mentioned earlier.

However, we definitely shouldn't approach this to start. Just documenting my thoughts.

draft/XXXX-multipart-upload/README.md
mvandenburgh: It seems that the underscore in `_uploads` makes it an invalid database name -
https://www.arangodb.com/docs/stable/data-modeling-naming-conventions-database-names.html

Should we just go with `uploads` for the DB name then?

waxlamp: That should be fine.

draft/XXXX-multipart-upload/README.md
jjnesbitt: Since we've pivoted to a parallel upload approach, the metadata that we'd want to store would be a bit different regarding upload progress. Perhaps something like:

```
{
  "id": <uuid>,
  "collection": <string>,  # Collection that stores the upload data
  "filename": <string>,    # The name of the uploading file client-side (only used as a hint to the client) 
  "size": <int>,           # The size of the total file in bytes (as reported by the client)
  "chunk_size": <int>,     # The size of each chunk in bytes
  "progress": {
    [chunk_seq: int]: true
  }
}
```

The new thing here is the `progress` field, which would basically be a checklist for each chunk sequence. That is, when a chunk is uploaded successfully, the following is performed:
```
upload_doc["progress"][chunk_seq] = true
```

This means that on the client, to resume an interrupted upload, you would begin uploading all chunks who's chunk sequence (or index) is not in the `progress` field.

This also means that determining the upload progress percentage can be done by dividing the number of entries in the `progress` field by the number of chunks (which is `size`/`chunk_size`).

Let me know what you think.

JackWilb (approved):

jjnesbitt (approved):

PR #14, opened 2020-08-04
Strike status field from template and existing RFCs

waxlamp:
Do you think we need this field in the RFCs themselves? It seems like moving the files between directories is enough. What do people think?


JackWilb (approved):

jjnesbitt (approved):
I suppose I'm okay with it

PR #17, opened 2020-10-28
RFC: Add proposal for missing values in typed cols

### LINKED ISSUE #16: Create RFC for missing data in typed columns
We should allow columns with a specific type (e.g. number) to have missing values. This RFC will have to consider whether this is allowed in Arango

JackWilb:
Closes #16

### DIFF
--- a/draft/XXXX-missing-values-in-typed-cols/README.md
+++ b/draft/XXXX-missing-values-in-typed-cols/README.md
@@ -13,9 +13,7 @@ There will likely be some client- and server-side considerations for handling mi
 
 ## Motivation
 
-We now have a motivation to supply data schema information to visualization applications, UpSet. Upset requires set membership information and other data types to generate its visualization. We can also utilize this type information in the other visualization applications that support multinet (MultiLink and MultiMatrix).
-
-Since we now need to pass semantic typing into these applications, knowing what the types are, even if there is missing data, is essential. As such, we'll need to modify our typing logic to handle missing data. This stretches from the frontend, where we provide type suggestions, to the backend, where we store the type information and disseminate it to the vis applications.
+In general, we want to support typed columns for a variety of purposes, including visualization, comparison, grouping, etc. In the real world, datasets almost always have missing entries, and, as such, MultiNet will need to accommodate column typing with missing data. 
 
 ## Proposals
 
--- a/draft/XXXX-missing-values-in-typed-cols/README.md
+++ b/draft/XXXX-missing-values-in-typed-cols/README.md
@@ -29,7 +29,7 @@ For the other repos, `multinet-server` and `multinetjs`, the considerations are 
 
 ## Backwards Compatibility
 
-There will be no added backwards incompatibilities from this change. There may be some knock-on consequences for apps that can utilize types in the case that there was data uploaded with the old type definitions, but the impact will be minimal.
+There is a potential for some backwards incompatibilities due to the types of the data being changed. Currently, missing data is represented as an empty string; whereas, when we have typed data, the returned value will be `null`. This could impact some of our data processing in the client/vis apps.
 
 ## Reference Implementation
 
--- a/draft/XXXX-missing-values-in-typed-cols/README.md
+++ b/draft/XXXX-missing-values-in-typed-cols/README.md
@@ -9,7 +9,7 @@ Last-Modified: 2020-10-26
 
 As we move to having typed data in Multinet, it is important to handle missing data elegantly, even with typing. Currently, if a column has any amount of missing data, we assume that the data is a "label" type. Since the "label" type is essentially a catch-all for data that we can't provide a meaningful type for, it's important that if there is a semantically meaningful type, we provide it to visualization applications. This must, also, be true in the case where we have mostly complete data with just a few missing rows. Without this information, the client applications cannot adequately choose a visualization for these columns.
 
-There will likely be some client- and server-side considerations for handling missing data. This will include correctly parsing data and when reading it into the ArangoDB types (or we may choose to leave them as strings in the db). Types that we define, will be referenced as semantic types, and native ArangoDB types will be referenced as Arango types
+There will likely be some client- and server-side considerations for handling missing data. This will include correctly parsing data and when reading it into the ArangoDB types (or we may choose to leave them as strings in the db). Types that we define will be referenced as *semantic types*, and native ArangoDB types will be referenced as *Arango types*.
 
 ## Motivation
 
@@ -19,7 +19,7 @@ Since we now need to pass semantic typing into these applications, knowing what 
 
 ## Proposals
 
-We should provide type information for columns that are of one type, but that have missing values. There are at least 3 places that we should support this paradigm:
+Multinet should support typing of columns with missing entries, rather than just assuming such columns are `label` type. There are at least 3 places that we should support this paradigm:
 
 1. `multinet-client` on data preview/type suggestion.
 2. `multinet-server` on data upload and data fetching.
--- a/draft/XXXX-missing-values-in-typed-cols/README.md
+++ b/draft/XXXX-missing-values-in-typed-cols/README.md
@@ -7,7 +7,7 @@ Last-Modified: 2020-10-26
 
 # RFC XXXX: Missing Values In Typed Columns
 
-As we move to having typed data in Multinet, it is important to handle missing data elegantly, even with typing. Currently, if a column has any amount of missing data, we assume that the data is a "label" type. Since the "label" type is essentially a catch-all for data that we can't provide a meaningful type for, it's important that if there is a semantically meaningful type, we provide it to visualization applications. This must, also, be true in the case where we have mostly complete data with just a few missing rows. Without this information, the client applications cannot adequately choose a visualization for these columns.
+As we move to having typed data in MultiNet, it is important to handle missing data elegantly, even with typing. Currently, if a column has any amount of missing data, we assume that the data is a "label" type. Since the "label" type is essentially a catch-all for data that we can't provide a meaningful type for, it's important that if there is a semantically meaningful type, we provide it to visualization applications. This must, also, be true in the case where we have mostly complete data with just a few missing rows. Without this information, the client applications cannot adequately choose a visualization for these columns.
 
 There will likely be some client- and server-side considerations for handling missing data. This will include correctly parsing data and when reading it into the ArangoDB types (or we may choose to leave them as strings in the db). Types that we define will be referenced as *semantic types*, and native ArangoDB types will be referenced as *Arango types*.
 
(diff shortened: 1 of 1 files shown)


draft/XXXX-missing-values-in-typed-cols/README.md
```diff
@@ -0,0 +1,44 @@
...
+Author: Jack Wilburn
+Created: 2020-10-26
+Last-Modified: 2020-10-26
+---
+
+# RFC XXXX: Missing Values In Typed Columns
+
+As we move to having typed data in Multinet, it is important to handle missing data elegantly, even with typing. Currently, if a column has any amount of missing data, we assume that the data is a "label" type. Since the "label" type is essentially a catch-all for data that we can't provide a meaningful type for, it's important that if there is a semantically meaningful type, we provide it to visualization applications. This must, also, be true in the case where we have mostly complete data with just a few missing rows. Without this information, the client applications cannot adequately choose a visualization for these columns.
+
+There will likely be some client- and server-side considerations for handling missing data. This will include correctly parsing data and when reading it into the ArangoDB types (or we may choose to leave them as strings in the db). Types that we define, will be referenced as semantic types, and native ArangoDB types will be referenced as Arango types
```
waxlamp:
```suggestion
There will likely be some client- and server-side considerations for handling missing data. This will include correctly parsing data and when reading it into the ArangoDB types (or we may choose to leave them as strings in the db). Types that we define will be referenced as *semantic types*, and native ArangoDB types will be referenced as *Arango types*.
```

waxlamp: Eventually, when an RFC for the column types API appears (attn: @AlmightyYakob), we'll want to link to that RFC from here.

draft/XXXX-missing-values-in-typed-cols/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Motivation
+
+We now have a motivation to supply data schema information to visualization applications, UpSet. Upset requires set membership information and other data types to generate its visualization. We can also utilize this type information in the other visualization applications that support multinet (MultiLink and MultiMatrix).
+
+Since we now need to pass semantic typing into these applications, knowing what the types are, even if there is missing data, is essential. As such, we'll need to modify our typing logic to handle missing data. This stretches from the frontend, where we provide type suggestions, to the backend, where we store the type information and disseminate it to the vis applications.
```
waxlamp: I don't think this is the right motivation. It's more general: we want to support typed columns, and in the real world datasets almost always have missing entries, so we'll need to support that reality. This has very little to do with UpSet or any other specific vis application.

draft/XXXX-missing-values-in-typed-cols/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Proposals
+
+We should provide type information for columns that are of one type, but that have missing values. There are at least 3 places that we should support this paradigm:
```
waxlamp: I reworded this a bit; let me know what you think.

```suggestion
Multinet should support typing of columns with missing entries, rather than just assuming such columns are `label` type. There are at least 3 places that we should support this paradigm:
```

draft/XXXX-missing-values-in-typed-cols/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+1. `multinet-client` on data preview/type suggestion.
+2. `multinet-server` on data upload and data fetching.
+3. `multinetjs` on data upload and data fetching.
+
+For the type suggester, there is a code chunk in the [Reference Implementation](#reference-implementation) section that suggests a way that we can check for the type of a column, even when data is missing.
+
+For the other repos, `multinet-server` and `multinetjs`, the considerations are slightly different. We'll need to think about how to store and serve data with missing values, and how to disseminate the type information given that there are missing values. Additionally, an open question is, should we tell client applications to expect some missing data in typed columns? This could help when rendering visualizations.
```
waxlamp: My thoughts on this:
1. Storing/serving missing values is easy: this is the exact semantics of the JSON/JavaScript `null` value, so we simply store `null` in the DB, and serve it upon retrieval just like any other value.
2. Whether to indicate missing values in a column is a slightly harder question, but I'm leaning towards "yes". If we're considering the type of a column to be a special kind of column metadata entry, then we could also have a boolean flag called `complete` or something like that (we will need to think hard about what term is best) as well. As with all other metadata, vis apps would be free to make use of or ignore this flag at its discretion.

draft/XXXX-missing-values-in-typed-cols/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Backwards Compatibility
+
+There will be no added backwards incompatibilities from this change. There may be some knock-on consequences for apps that can utilize types in the case that there was data uploaded with the old type definitions, but the impact will be minimal.
```
waxlamp: I think technically there _will_ be a backwards incompatibility, since this proposal (and the upcoming one for typed column storage in the first place, which this depends on) will change how data is stored in the DB. That is, any app that was relying on numeric data from Multinet will no longer be receiving strings representing the data that they must convert themselves. The oddity here is simply that we don't really have any such apps currently.

In any case, I think this backwards incompatibility for *this* RFC in particular is just that apps may now receive `null` when before they would always receive something non-`null`.

draft/XXXX-missing-values-in-typed-cols/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Reference Implementation
+
+The client already has logic to calculate the type of the columns but does not allow for missing values as it stands. This implementation should just be an extension of the existing logic.
+
+Jake gave a suggestion for how we might calculate if a column is actually the correct type with some missing values:
+
+```
+const number = entry.number + entry.empty === entry.total;
+```
```
waxlamp, line 44: Beyond this, we can also display something when `entry.empty > 0` to indicate that the column has missing values (if the uploader intended for this not to be the case, it would be a useful prompt to look at their raw data again).

draft/XXXX-missing-values-in-typed-cols/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+For the type suggester, there is a code chunk in the [Reference Implementation](#reference-implementation) section that suggests a way that we can check for the type of a column, even when data is missing.
+
+For the other repos, `multinet-server` and `multinetjs`, the considerations are slightly different. We'll need to think about how to store and serve data with missing values, and how to disseminate the type information given that there are missing values. Additionally, an open question is, should we tell client applications to expect some missing data in typed columns? This could help when rendering visualizations.
+
+## Backwards Compatibility
+
+There will be no added backwards incompatibilities from this change. There may be some knock-on consequences for apps that can utilize types in the case that there was data uploaded with the old type definitions, but the impact will be minimal.
+
+## Reference Implementation
```
waxlamp: This is not really a reference implementation. I'd change this header to "Implementation Strategies" or something like that.

waxlamp (commented):
Good stuff, @JackWilb! Let's open some discussion in the comments and edit as needed before accepting this.

You may want to invite Alex's review on this too (or you could wait until we've hashed out a few things before getting his eyes on it).

draft/XXXX-missing-values-in-typed-cols/README.md
jjnesbitt: I'm not necessarily against storing the "completeness" of a column, but is there an example use case for this?

waxlamp: I don't think we have a concrete use case now, but the idea would be to signal to apps that they may need to check for `null` when doing computations, etc. If setting a metadata flag to indicate this has problems associated with it, we don't have to do it, but I can see it streamlining vis app code. For instance, check to see if `!metadata.complete`, and run an imputation or filtering step on the data before proceeding.

Not a big deal if we don't include it, or if we do so after a concrete need arises.

JackWilb: I think the client should be deciding if there are `null` values in the dataset instead of relying on the server for that info for now. Unless that become a bottleneck, at which point we can revisit this

draft/XXXX-missing-values-in-typed-cols/README.md
JackWilb: I like this idea. I'll make sure I note it down in the RFC

draft/XXXX-missing-values-in-typed-cols/README.md
JackWilb: Updated in 2872e82

draft/XXXX-missing-values-in-typed-cols/README.md
JackWilb: Updated in ad86218

draft/XXXX-missing-values-in-typed-cols/README.md
JackWilb: Updated in 4493627

draft/XXXX-missing-values-in-typed-cols/README.md
JackWilb: Updated in a22ca43

draft/XXXX-missing-values-in-typed-cols/README.md
waxlamp: Just FYI, that commit does not seem to address my original comment.

draft/XXXX-missing-values-in-typed-cols/README.md
waxlamp: You're right. Even if the type indicates there are no missing values, I don't expect that to make coding vis apps any easier: you would still need to check for `null` in some branch of the code where the type indicates there are missing values.

Together with @AlmightyYakob's argument that we don't have a valid need for this right now, that means we shouldn't support this now.

Go ahead and update the paragraph to reflect that, plus my comments about using `null` values above.

draft/XXXX-missing-values-in-typed-cols/README.md
waxlamp: We should aim for decisiveness in these RFCs--we want to use them as a guide to implementation after they're settled. So if these feature requires further debate, let's do it and decide one way or the other as much as possible.

In this case, I agree that the client could check for missing data itself and use that fact to present an indicator in the UI. The API may not need to be involved at all. If you feel the same way, go ahead and resolve the language to reflect that.

(Sometimes it may really be impossible to decide until implementation time how something should be, and in those cases of course it's ok to make a note of that in the RFC. But this doesn't seem to be one of those times.)

draft/XXXX-missing-values-in-typed-cols/README.md
waxlamp: Ah I think you meant 535593f6a7ca1666dd52d51928dfe0d130413c36. Looks good, thanks.

JackWilb: Whoops, yes

draft/XXXX-missing-values-in-typed-cols/README.md
JackWilb: 0d938f8

draft/XXXX-missing-values-in-typed-cols/README.md
JackWilb: 0d938f8

waxlamp (approved):

jjnesbitt (approved):

PR #18, opened 2020-10-29
RFC: Table Metadata

jjnesbitt:


draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+The main motivation for this is storing type information for each column in a table. We now have at least two use cases for storing this information.
+
+1. Basic computation/comparison. If we attempt to use these tables for creating graphs and deriving useful data, it's necessary that we store type information, so that we are able to properly compare and compute certain values. For example, if you have a column of numbers, but we store them as strings (the current behavior), then if you try to do comparisons of these column values between different entries, you either will be doing incorrect string comparisons, or will simply not be able to do the comparisons/computations you need to.
+
+2. Providing this information to other applications, so that they do not have to do this data inspection themselves, or if this information is required by those applications. Specifically, [UpSet](https://github.com/VCG/upset) is the first application for which this will be used.
+
+## Proposals
+
+#### Data Model
```
waxlamp, line 23:
```suggestion
### Data Model
```

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+There are several way we could go about this, but in my opinion, the cleanest way is to store this information in a special collection, in the workspace that the table resides in. This special collection would be named something like `multinet_meta`, and each entry would correspond to the metadata on a specific graph or table. I think the data model would look something like this:
```
waxlamp:
```suggestion
There are several ways we could go about this, but in my opinion, the cleanest way is to store this information in a special collection, in the workspace that the table resides in. This special collection would be named something like `multinet_meta`, and each entry would correspond to the metadata on a specific graph or table. I think the data model would look something like this:
```

waxlamp: It's ok not to hedge your statements with "I think" and "in my opinion", etc. The RFC process will bash out the best consensus, which we'll then want to state in declarative terms.

```suggestion
There are several way we could go about this, but the cleanest way is to store this information in a special collection, in the workspace that the table resides in. This special collection would be named something like `multinet_meta`, and each entry would correspond to the metadata on a specific graph or table. The data model would look something like this:
```

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+Ideally, our API would allow for the creation of a table/graph record, with more versatile options for specifying metadata. However, our API isn't currently structured this way. Instead, in order to conform to our existing API, we would likely just add this metadata object as a query parameter onto each upload endpoint. For retrieving this metadata, the API should resemble the following:
+
+```
+GET /api/workspaces/{workspace}/tables/{table}/metadata
+```
+
+##### Specifying Metadata on Table Creation
+
+The data model of the API would differ slightly from that of what we store internally, to allow for a single request to describe metadata on multiple entities (mainly in the case of graphs. The data format of this API may resemble the following:
```
waxlamp:
```suggestion
The data model of the API would differ slightly from that of what we store internally, to allow for a single request to describe metadata on multiple entities (mainly in the case of graphs). The data format of this API may resemble the following:
```

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+    ]
+  }
+  graph?: {                    // Only present if the major type is "graph"
+    // TBD
+  }
+
+}
+```
+
+#### API
```
waxlamp, line 47:
```suggestion
### API
```

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+1. Specifying the metadata on table creation
+2. Retrieving this metadata
+
+Ideally, our API would allow for the creation of a table/graph record, with more versatile options for specifying metadata. However, our API isn't currently structured this way. Instead, in order to conform to our existing API, we would likely just add this metadata object as a query parameter onto each upload endpoint. For retrieving this metadata, the API should resemble the following:
+
+```
+GET /api/workspaces/{workspace}/tables/{table}/metadata
+```
+
+##### Specifying Metadata on Table Creation
```
waxlamp, line 59:
```suggestion
#### Specifying Metadata on Table Creation
```

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+        key: str               // E.g. {"foo": "bar"}, "foo" is the key
+        type: "number" | "label" | "category" | "date" | "bool"
+      }
+    ]
+  }
+  graph?: {                    // Only present if the major type is "graph"
+    // TBD
+  }
+
+}
```
waxlamp: Do we need to store the object type (table vs network) and then store metadata under differently named keys? Why not just dump `item_type` and then replace the `table`/`graph` entries by `metadata`?

draft/XXXX-table-metadata/README.md
waxlamp: I'm not sure why we would need to update metadata on multiple things at once. Generally, tables are uploaded, then networks are created by selecting tables. At the time a network is created, I can see having some metadata also set up to be set on it. If you're thinking of the case where a data format automatically creates tables and a network, even in that case it seems like multiple calls to set metadata on different objects would be simpler to implement. What do you think?

waxlamp (commented):
Thanks for this, @AlmightyYakob.

One general comment here: we're not supposed to call them "graphs", but rather "networks".

draft/XXXX-table-metadata/README.md
AlmightyYakob: I actually prefer to use h4 in these sections, as to me it's harder to distinguish an h3 from an h2 as opposed to using an h4; I think it reads better.

draft/XXXX-table-metadata/README.md
jjnesbitt: The main reason was to contain the likely different schemas under a possibly absent key, so that the definition doesn't contain a bunch of possibly absent keys at the top level.

I do think it's good to keep track of the type in each document (table vs graph), but I suppose it might not be strictly necessary, as the keys would imply this.

draft/XXXX-table-metadata/README.md
jjnesbitt: With our current API, we would need to pass the type information of both node and edge tables along at graph creation time, since the upload/creation is a synchronous call, and it needs this information _before_ creating the tables/graphs, as this will affect the data type that is chosen to store each column.

waxlamp: But in general, the tables are created first, then a network is created. The type information would already be there in the table metadata. Right?

But I think it could be different for the special format network creation routines, like the ones for nested JSON and D3-formatted graph data. Is that what you're referring to here?

jjnesbitt: Yes, this is specifically for uploading a graph directly (D3 JSON, Nested JSON or Newick). In these cases, the tables (and subsequent graph) will be created synchronously in the api call, so we would need to provide this metadata as a part of the API call which creates them.

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Motivation
+
+The main motivation for this is storing type information for each column in a table. We now have at least two use cases for storing this information.
+
+1. Basic computation/comparison. If we attempt to use these tables for creating graphs and deriving useful data, it's necessary that we store type information, so that we are able to properly compare and compute certain values. For example, if you have a column of numbers, but we store them as strings (the current behavior), then if you try to do comparisons of these column values between different entries, you either will be doing incorrect string comparisons, or will simply not be able to do the comparisons/computations you need to.
+
+2. Providing this information to other applications, so that they do not have to do this data inspection themselves, or if this information is required by those applications. Specifically, [UpSet](https://github.com/VCG/upset) is the first application for which this will be used.
```
JackWilb:
```suggestion
2. Providing this information to other applications, so that they do not have to do this data inspection themselves, or if this information is required by those applications. Specifically, [UpSet2](https://github.com/visdesignlab/upset2) is the first application for which this will be used.
```

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Proposals
+
+#### Data Model
+There are several ways we could go about this, but in my opinion, the cleanest way is to store this information in a special collection, in the workspace that the table resides in. This special collection would be named something like `multinet_meta`, and each entry would correspond to the metadata on a specific graph or table. I think the data model would look something like this:
```
JackWilb: Could there be conflicts if someone uploaded a table called `multinet_meta` since this new collection will reside in the workspace?

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+        key: str               // E.g. {"foo": "bar"}, "foo" is the key
+        type: "number" | "label" | "category" | "date" | "bool"
+      }
+    ]
+  }
+  graph?: {                    // Only present if the major type is "graph"
+    // TBD
+  }
+
+}
```
JackWilb: How do we store multiple of these for a workspace, can we just append more to the collection for each table/network?

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+#### API
+There are two main operations that are necessary for this functionality.
+
+1. Specifying the metadata on table creation
+2. Retrieving this metadata
```
JackWilb: Should we consider updating the metadata to be a critical function, too?

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+{
+  graph?: {
+    // TBD, any relevant graph metadata
+  }
+  tables: [  // For a single CSV, this array must contain only one item.
+    {
+      // Unchanged from the table field above
+    }
+  ]
+}
```
JackWilb: This looks like it only allows for one network. What if we want the metadata from the whole workspace?

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+However, handling the case of graphs may not be necessary in the first approach. Feedback on how to best implement this is welcome.
+
+## Backwards Compatibility
+
+All changes may be backwards compatible to the server, as this new query parameter would ideally be an "optional" parameter. However, if it's desired that this field be mandatory, then this breaks backwards compatibility for all upload endpoints.
```
JackWilb: I think this should be mandatory, because some visualizations will work best when it's specified. I'd like to hear some counterarguments, though, if you guys have some

JackWilb (commented):
Generally, this looks great. I just had a couple of comments and areas that I think you can expand on in the RFC

draft/XXXX-table-metadata/README.md
jjnesbitt: Yes, but in that case, we would just disallow a table to be named `multinet_meta`, or whatever we land on for the name.

draft/XXXX-table-metadata/README.md
jjnesbitt: Yes, for any given workspace, the metadata collection would store one document per table/graph.

draft/XXXX-table-metadata/README.md
jjnesbitt: Yes, I can add that.

draft/XXXX-table-metadata/README.md
jjnesbitt: This RFC focuses on metadata for entities in a workspace (graphs and tables). If we want to store metadata on an entire workspace, the place to do that would be the `workspace_mapping` collection, which is already a misnomer, as it's really a workspace metadata collection.

draft/XXXX-table-metadata/README.md
jjnesbitt: To me it feels a little dicey to _require_ that type information (generally speaking metadata) be specified when creating an entity. In my opinion, these should act as "type hints", and we should have some default behavior.

draft/XXXX-table-metadata/README.md
JackWilb: In the client we could have a nice message that it's a reserved tablename. I think `_multinet_metadata` might be best? It also would then indicate that it's private/not supposed to be accessed

jjnesbitt: Unfortunately, table names can't begin with an underscore or number 

https://www.arangodb.com/docs/stable/data-modeling-naming-conventions-collection-and-view-names.html

draft/XXXX-table-metadata/README.md
waxlamp: Are we planning to restrict what metadata can be present for a table or network? I was thinking this could be somewhat like item metadata in Girder (which doesn't really have any schema at all) or something with "recommended" properties like `description` but with others being arbitrary key/value pairs. If we think of it that way, I don't think you would need the `table` or `graph` keys in the top level, just a generic `metadata`, regardless of whether it's a table or network. The other top-level keys can be useful stuff like `name`, `type`, `columnTypes` (for tables), etc.

Hopefully that clarifies what I'm thinking.

draft/XXXX-table-metadata/README.md
waxlamp: We'll need to figure out UI/UX for pre-analyzing the data tables and allowing the user to specify types. But your approach makes sense to me now.

draft/XXXX-table-metadata/README.md
waxlamp: Unless I'm forgetting something, I don't think this will cause a problem. We don't store tables as collections with the user-specified names; we take their name for the table and stick it in an indirection table that maps to a UUID-looking collection name. So we can still maintain a collection called `multinet_meta` or `metadata` or whatever we like, and still maintain tables with that same name.

Right?

draft/XXXX-table-metadata/README.md
waxlamp: I largely agree with @AlmightyYakob on this. But de facto it will be mandatory in a sense: since we expect the vast majority of people to upload data to Multinet through the client, then whatever the client does will appear to be the normal course of action. We already have the type specification UI built for CSV uploads; this means that when all the pieces are connected, _all_ CSV uploads will include type information. Maybe in some future revision we can decide to make this mandatory, but I don't see a need for it now.

If there's no type information for a particular column, I think it would make sense to implicitly treat that column as `label` (since this is effectively the current semantics of typeless upload anyway).

draft/XXXX-table-metadata/README.md
jjnesbitt: We only do this indirection for storing the name of workspaces (as currently they're the only thing we allow to be renamed). For tables/graphs, their names are stored verbatim.

waxlamp: Oh shoot, I *was* forgetting something!

I'm fine with just disallowing certain table names by failing during validation (or even just in the form where we supply the table name during upload). We could do name indirection for tables and networks but that would get pretty hairy and for not a lot of benefit.

draft/XXXX-table-metadata/README.md
jjnesbitt: There's a difference here between "required" and other metadata. What I've outlined above is "required" in the sense that we need that information in order to properly store the uploaded items, and provide the type information to external APIs in a structured way. I'd say if we want to add general metadata, that could go under a `"meta"` key or something, so that arbitrary metadata isn't mixed in with required metadata at the top level.

draft/XXXX-table-metadata/README.md
jjnesbitt: 70e1745

draft/XXXX-table-metadata/README.md
jjnesbitt: Yeah, I'd say that for any missing columns, it is assumed to be label, and just stored as string. This would also apply for all metadata, so if no metadata is passed along, all columns are just treated as strings/labels.

draft/XXXX-table-metadata/README.md
JackWilb: Let's make an issue for this in the client repo as the RFC is settled

waxlamp: @AlmightyYakob, so it turns out we can use a [system collection](https://www.arangodb.com/docs/stable/http/collection-creating.html) for this, right?

draft/XXXX-table-metadata/README.md
waxlamp: Ok I think I understand. Let's go with your design then, since I'm having trouble explaining why I feel weird about this `table`/`graph` split. If that remains a problem in the future, we can always adjust if necessary.

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+2. Providing this information to other applications, so that they do not have to do this data inspection themselves, or if this information is required by those applications. Specifically, [UpSet2](https://github.com/visdesignlab/upset2) is the first application for which this will be used.
+
+## Proposals
+
+#### Data Model
+There are several way we could go about this, but the cleanest way is to store this information in a special collection, in the workspace that the table resides in. This special collection would be named something like `multinet_meta`, and each entry would correspond to the metadata on a specific graph or table. The data model would look something like this:
```
waxlamp: I changed the language to talk about system collections:

```suggestion
There are several way we could go about this, but the cleanest way is to store this information in a special [system collection](https://www.arangodb.com/docs/stable/http/collection-creating.html), in the workspace that the table resides in. This special collection would be named `_multinet_meta`, and each entry would correspond to the metadata on a specific graph or table. The data model would look something like this:
```

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,84 @@
...
+  }
+  graph?: {                    // Only present if the major type is "graph"
+    // TBD
+  }
+
+}
+```
+
+#### API
+There are two main operations that are necessary for this functionality.
```
waxlamp:
```suggestion
There are three main operations that are necessary for this functionality.
```

draft/XXXX-table-metadata/README.md
```diff
@@ -0,0 +1,84 @@
...
+  }
+  tables: [  // For a single CSV, this array must contain only one item.
+    {
+      // Unchanged from the table field above
+    }
+  ]
+}
+```
+
+However, handling the case of graphs may not be necessary in the first approach. Feedback on how to best implement this is welcome.
```
waxlamp: Yeah, let's not worry about network metadata in the current iteration. If you want, you can just punt on that in this text:

```suggestion
In this iteration, network metadata will not be implemented.
```

draft/XXXX-table-metadata/README.md
waxlamp: You can edit this section to say that there won't be any breakages.

waxlamp (changes requested):
In addition to the suggestions, you need to change all instances of the word "graph" to "network" (that's the official terminology for Multinet). But I think this is pretty much ready to go.

draft/XXXX-table-metadata/README.md
jjnesbitt: What do you think about calling the collection `_metadata`? I think that reads better.

waxlamp: I like it.

draft/XXXX-table-metadata/README.md
jjnesbitt: Yes

draft/XXXX-table-metadata/README.md
jjnesbitt: b2409bf

waxlamp (approved):

JackWilb (approved):

PR #22, opened 2021-01-19
RFC: Robust Upload System

### LINKED ISSUE #20: Write new S3 upload API RFC
@waxlamp and I have discussed implementing a new mechanism for upload, which involves using S3. The main benefit to this is the ability for long running uploads, saved progress, larger files, etc.

jjnesbitt:
This is the long await "S3 API RFC". Sorry for the delay, but I believe this is in a good spot now.
Fixes #20


draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+    <ETag returned from the first signed url>,
+    <ETag returned from the second signed url>,
+    ...
+  ]
+}
+```
+
+**NOTE:** Similar to the `uploadPartUrls` array mentioned in the Upload creation step, the ordering of the `uploadPartTags` array is significant.
+
+At this point, the sever downloads the reconstructed file from S3, and performs data processing, table/graph creation, etc. with the appropriate processor for that data type. This process doesn't technically need to be done asynchronously, as if the data is small enough it's feasible that this could be completed within the time frame of a single request. However, in the interest of scalability and robustness, this HTTP request should initiate an asynchronous task to process the data.
```
waxlamp: Does "asynchronous task" mean introducing Celery or a similar system? I think you can avoid hedging here and just assert that this part is done asynchronously, since that's sort of the thrust of this whole effort.

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+1. Read in data until a fully formed object can be parsed (a table row, for example).
+2. Run validation on this object
+3. If validation is passed, insert this object into the target location. If validation fails, save the validation error, and either continue processing more data, or halt data processing.
+4. Repeat steps 1-3 until all data has been processed
+5. If there are errors resulting from this processing, delete the partially formed tables/graphs, and return the validation errors to the user.
+
+These errors would likely be stored in the upload model (yet to be outlined), however, it may make sense to introduce another model that would track this status more specifically. Feedback is welcome here.
```
JackWilb: We need to design this to complete within the Heroku limitations of 30 seconds per part, too. Although 1 row probably wouldn't cause us any problems

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+
+### Tracking upload progress
+
+The current proposal requires that the client upload all parts, and then returns the ETags of these parts all at once when finalizing the upload. However, this poses a couple of possible pitfalls:
```
JackWilb: Another is, what if one part fails out of 1000? That would introduce a lot of error checking logic on the client side so I don't think your first concept is reliable enough. This second idea, to use the `finalizePart` endpoint seems much more reliable

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+**NOTE:** The ordering of the `uploadPartUrls` array is significant, as it represents the specific URL that is to be used to upload each corresponding part. These URLs may be called concurrently, but the payload for each URL must be the section of the file which corresponds to its index.
+
+
+### Upload the individual parts
+
+This is a client side operation, and involves making an HTTP `PUT` request to each URL in the `uploadPartUrls` array. The payload for each of these requests is the section of the file data which corresponds to the index of the signed URL in the array (i.e. if the file is 100MB total, and the `uploadPartUrls` array contains 5 entries, the payload for the first request will be the bytes 1 - 20,000, the payload for the second request will be bytes 20,001 - 40,000, etc.). The response for each `PUT` call is an `ETag`, which must be recorded respective to the part index and used in the finalization call later.
+
+
+### Finalize the upload
```
JackWilb: We need to flesh this out with some possible return values. I can see a point where the client issues a finalize before the data has settled or while some is missing. How would we handle that on the API and the client?

draft/XXXX-robust-api/README.md
JackWilb: Or, as mentioned above, we could use async for this

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+  ],
+  "uploadPartTags": [ // This will be an empty list initially
+    "<ETag from the first uploaded part>",
+    "<ETag from the second uploaded part>",
+    ...
+  ],
+  "metadata": {
+    // metadata like type information goes here
+  }
+}
```
JackWilb: We might want to compute a hash so that we can check if the upload was already started. If it was, we can resume instead of starting over again

JackWilb (commented):
Here are some preliminary thoughts. 

I think that your second idea of tracking parts instead of the whole file as one is much more robust and gives us the features that we'd like (resuming, progress tracking, etc.). If we're all in agreement, it'd be great to see a second draft with that idea in more detail.

draft/XXXX-robust-api/README.md
jjnesbitt: Yeah imo this should be done asynchronously using something like celery, even though for small enough data we could probably get away with it.

draft/XXXX-robust-api/README.md
jjnesbitt: I'm a bit confused, what error checking logic would need to be introduced? If any of the upload parts fails, I'd think the client would just declare the upload as failed. Regardless, I do agree the second approach is probably more flexible.

JackWilb: I meant if the client sends a finalize without all the parts being done (or one failing). Then the server would error when it tries to read the data in. The solution would be to set the client logic to not send the finalize, but that seems less robust than the server keeping track and rejecting a finalize until all the parts have settled.

Error checking was a bit of a misnomer, because it has a specific meaning in this context. I meant just checking for a missed upload part

jjnesbitt: > I meant if the client sends a finalize without all the parts being done (or one failing). Then the server would error when it tries to read the data in. The solution would be to set the client logic to not send the finalize, but that seems less robust than the server keeping track and rejecting a finalize until all the parts have settled.

My proposal was actually _not_ to have the client send the finalize request, but rather, have the server send the request. All the client would need to do is to send the appropriate `Etags` on upload part completion. I believe this split between access to upload parts and access to the entire upload object can be enforced through permissions on the S3 object.

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
+---
+RFC: XXXX
+Author: Jacob Nesbitt
+Created: 2020-06-29
+Created: 2020-12-12
```
waxlamp:
```suggestion
```

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+Last-Modified: 2020-12-12
+---
+
+# RFC XXXX: RFC Template
```
waxlamp: Can you update the title?

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+# Proposal
+
+## Upload
+Currently, our upload system deals with upload, data processing, and table creation, all at once and in the same request. Since we deploy on heroku, and we need to keep our requests under 30 seconds (as we should regardless), this severely limits both the size of the data we can upload. In order to address this, we can make use of Amazon S3. The new upload process would follow this rough path:
+
+
+### Create an upload
+
+`POST /api/workspaces/{workspace}/upload/create`
```
waxlamp: `POST` implies `create` so you should leave that off the endpoint route. (Note also that our REST API uses plurals in naming resources within a route.)

```suggestion
`POST /api/workspaces/{workspace}/uploads`
```

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+Create an upload, returning an `Upload` object that (among other things) will contain a list of pre-signed URLs, that are to be used for uploading each part of the file to S3. **Note:** This means that the amount of parts / size of each part is determined by the server. The parameters for this call will include the following
+
+* `dataType` - The data type of the file to upload, which must be one of the supported (or installed, if using a plugin system) data types. E.g. `csv`, `d3_json`, `nested_json`, `newick`, etc. If this parameter does not match one of the supported data types, a `4XX` error is thrown.
```
waxlamp: I wonder if this should be passed on *finalize*, rather than upload creation, *unless* we want to validate the data before it's uploaded (which seems to be almost a necessity for very large files).

I also wonder if we should create offline tools for validating preflight data... It would be easier to write and run that sort of validator as, e.g., a Python script after you've prepped your data but before you even bring it near Multinet. This notion of "preflight validation" is something we should discuss.

draft/XXXX-robust-api/README.md
waxlamp: Resumption wouldn't be through this `POST` endpoint, though. The `_id` of the response would be used for that, I think. Upload interruption/resumption would have to be handled through a specialized system that would have a UI in the client.

I agree that a hash would be useful in verifying that a resumed upload is occurring with the same file that was interrupted before, though. On the other hand, I'm sure there are best practices for this kind of thing and we should find out what those are.

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+```
+
+**NOTE:** The ordering of the `uploadPartUrls` array is significant, as it represents the specific URL that is to be used to upload each corresponding part. These URLs may be called concurrently, but the payload for each URL must be the section of the file which corresponds to its index.
+
+
+### Upload the individual parts
+
+This is a client side operation, and involves making an HTTP `PUT` request to each URL in the `uploadPartUrls` array. The payload for each of these requests is the section of the file data which corresponds to the index of the signed URL in the array (i.e. if the file is 100MB total, and the `uploadPartUrls` array contains 5 entries, the payload for the first request will be the bytes 1 - 20,000, the payload for the second request will be bytes 20,001 - 40,000, etc.). The response for each `PUT` call is an `ETag`, which must be recorded respective to the part index and used in the finalization call later.
```
waxlamp: This part is all via the S3 API, right? (Just trying to make sure I understand how this all works.)

draft/XXXX-robust-api/README.md
waxlamp: How would that happen? Doesn't the client need to wait until it gets a 200 response from every part upload before it would issue a finalize call? And, if it did issue that call before all data was settled, the server could respond with 400 because it would see that some of the ETag values are missing.

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+There are changes in addition to the endpoints mentioned above, which may not be crucial to the client, but should be detailed.
+
+### Upload Collection
+
+Each workspace should contain its own upload collection, which will be used to store each upload document shown above. This will need to be accompanied by a new `Upload` ORM, which will facilitate operations on uploads.
+
+### Additional Upload Endpoints
+
+`GET /api/workspaces/{workspace}/upload/{id}`
```
waxlamp:
```suggestion
`GET /api/workspaces/{workspace}/uploads/{id}`
```

draft/XXXX-robust-api/README.md
waxlamp: That everything isn't "small enough data" is precisely the problem we're trying to solve. Go ahead and make the assumption that we need async task support for this.

draft/XXXX-robust-api/README.md
waxlamp: I'm not sure if I understand the problem: for the client to finalize, it has to have all of the response ETags, which means all the part uploads succeeded. Right?

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+{
+  "partNumber": integer,
+  "ETag": string
+}
+```
+
+This would add to the `uploadPartTags` array on the upload document, which would allow for tracking the progress of this upload. After all file parts are uploaded, the upload finalize call is made as usual, but in this new approach, it would not contain any payload.
+
+
+Feedback on this suggested feature is appreciated.
```
waxlamp: The simpler approach is to have the client save the ETags in local storage. That would cover the common case, which is that an upload breaks in the user's browser, and the user fixes it from the same browser.

Recording successful part uploads by informing the server of each one is a more robust solution, but seems like it will require further design. It is an enhancement to the current proposal, so TBH I would say don't worry about it right now, and defer work on resumable uploads in a future push.

draft/XXXX-robust-api/README.md
```diff
@@ -0,0 +1,146 @@
...
+
+## Backwards Compatibility
+
+This RFC is not backwards compatible with the existing API, as the structure, upload procedure, etc. are all breaking changes.
```
waxlamp: To fix the breakage, we'll need to update the main client, and change the multinetjs functions related to uploads. Any other changes?

Please record the list of what needs to be changed in this section.

waxlamp (changes requested):
Some minor requests to complete the RFC metadata, and some questions/opinions.

This actually looks about ready to go to me, once the open threads are resolved to everyone's satisfaction.

There is a lot to do for this feature, so after this is finalized, let's create an action plan and figure out how to divide up the work. Thanks @AlmightyYakob!

draft/XXXX-robust-api/README.md
jjnesbitt: Yes, it would be celery. I can update this part to remove the uncertainty.

draft/XXXX-robust-api/README.md
jjnesbitt: Good catch, not sure why that was there.

draft/XXXX-robust-api/README.md
jjnesbitt: Yes. From the multipart upload docs:

> Your complete multipart upload request must include the upload ID and a list of both part numbers and corresponding ETag values. Amazon S3 response includes an ETag that uniquely identifies the combined object data. 

Which we would do through an Python S3 driver (`boto3`).

draft/XXXX-robust-api/README.md
jjnesbitt: Yes, ideally a client wouldn't have permissions to issue the finalize request (it would have to go through us). And if the client tried to issue a finalize request without all of the ETags, we would answer with a 4XX response, stating that they need to upload all ETags before finalizing the upload.

draft/XXXX-robust-api/README.md
jjnesbitt: I'm a bit confused. The only reason I could see us wanting to store a hash here is to hash the original file before uploading, which could then be used if the client is refreshed or the connection is broken, to ensure that the file they select to resume the upload exactly matches the original file.


In my opinion this is overkill, and I don't think we should worry about this to start. Adding a `filename` and `size` field to this metadata should be sufficient, and would allow the user to identify which file it is they need to select to resume upload.

As far as how the user would know which uploads they have in progress, there should be an endpoint of the form

```
GET /api/users/uploads
```

which would return a list of the uploads that the currently logged in user initialized. This is currently lacking from the RFC, so I'll ad that back in.

draft/XXXX-robust-api/README.md
jjnesbitt: As far as passing it on finalize vs on upload creation, I don't see a particular reason to pass it on finalize, since you'll know the data type at upload selection time. IMO, if we know this information we should pass it onto the API immediately, rather than wait.

I think the thoughts on offline tools are tangent to this issue. When it comes to our API, we shouldn't allow invalid data (from any source) to be inserted into our system. The only real way to do this is to validate the data in the API itself.

draft/XXXX-robust-api/README.md
jjnesbitt: Yes, these calls are being made to S3, and need to adhere to their multipart upload API design. This would also likely be done using an JS/TS API wrapper.

draft/XXXX-robust-api/README.md
jjnesbitt: There's two ways I can see this, as far as simplicity.:

1. The original proposal is simpler, because you're sending less requests back to the server, and storing things locally. However, that does mean you're storing two versions of the upload progress between the client/server.
2. The second approach is simpler, because it consolidates storage into the server, and the client isn't responsible for keeping track of everything, just doing its part to return it's answer to the server. However, that does involve more (albeit lightweight) requests.

You could go either way on these, but as far as API design, I'm more in favor of the second option. This is because if you _were_ to go with the first option, the client would be "keeping" information from the server. That is, if (for example) the browser is closed halfway through the upload, the API has no idea how far along the upload is, just that it's been started, and not finished. If that local storage was cleared for some reason, the client would have to re-upload all of those uploaded parts again. I know this is fairly unlikely to happen, but some people care about what information websites store in their browser etc., and would maybe delete this somehow. If we store this information on the server, it's nobody's burden to track except our own.

Let me know what you think.

draft/XXXX-robust-api/README.md
jjnesbitt: As far as I can see, the break in backwards compatibility are as follows:

- API
	- Uploads
	- Data Processing
- Client
	- Upload UI
- MultinetJS
	- Uploads

Am I missing anything?

draft/XXXX-robust-api/README.md
waxlamp: > As far as passing it on finalize vs on upload creation, I don't see a particular reason to pass it on finalize, since you'll know the data type at upload selection time.

The only reason would be to prevent accidents where you specified the wrong format at that early stage. For instance, if you upload a massive CSV file, but you mistakenly tell the API that it's a Newick file or something like that.

Basically, my thought process is that uploading a file to S3 can be done without knowing what type of file it is, so to require the minimum amount of information to do so would mean deferring the file format info to a later time.

> IMO, if we know this information we should pass it onto the API immediately, rather than wait.

But why? The server has to cache this information to make use of it at the right time. Furthermore, if you know the information now, then you'll know it later (when it's needed) as well.

I don't think this decision should hinder the initial implementation, so go with what you think is best (but consider my arguments in case they're persuasive). If we do it one way and decide it's done better another way, we can change it later. The discussion is valuable though. IMO, the fact that we're debating when to specify the file type points to a weakness in our overall design: I more and more am doubtful that dealing with file formats is something the server should be doing at all, but that's a totally separate topic of discussion.

> 
> I think the thoughts on offline tools are tangent to this issue. When it comes to our API, we shouldn't allow invalid data (from any source) to be inserted into our system. The only real way to do this is to validate the data in the API itself.

Agreed. My tangent thinking, as you put it, was musing on whether we can prevent disasters (i.e., having to re-upload massive files with minor mistakes corrected) by providing extra, preclient tooling. On that view, the API-side validation is to gate bad data from getting in, while the preflight validation would be to save the client frustration. I will file an issue around that and we can drop discussion of it for now.

draft/XXXX-robust-api/README.md
waxlamp: > I'm a bit confused. The only reason I could see us wanting to store a hash here is to hash the original file before uploading, which could then be used if the client is refreshed or the connection is broken, to ensure that the file they select to resume the upload exactly matches the original file.
> 
> In my opinion this is overkill, and I don't think we should worry about this to start. Adding a `filename` and `size` field to this metadata should be sufficient, and would allow the user to identify which file it is they need to select to resume upload.

No, you got it, this is just what I meant. The filename+size combo is a sort of hash, just one that doesn't prevent the file contents themselves from changing, but also one that doesn't require being stored somewhere.

I don't believe it's overkill to record the hash to prevent crazy (though rare) situations that would lead to corrupted data, but I agree with you that it's something we really don't need to do now. I'll file an issue so we don't forget about it.

> 
> As far as how the user would know which uploads they have in progress, there should be an endpoint of the form
> 
> ```
> GET /api/users/uploads
> ```
> 
> which would return a list of the uploads that the currently logged in user initialized. This is currently lacking from the RFC, so I'll ad that back in.

I think this `GET` endpoint should carry the same form as the `POST` endpoint you've proposed in this RFC. That one is targeted to workspaces, so perhaps there ought to simply be a `/api/uploads` section, and the model can record what workspace each upload is targeted to. Now that I'm thinking about it, I'm not sure the upload model even needs the workspace info, since the finalize call could be informed about which workspace to target the download to.

draft/XXXX-robust-api/README.md
waxlamp: I accept your argument.

My approach is only "simpler" in the sense that it requires less server code. It still requires (client) code, and it has the other non-simplicities you mentioned.

draft/XXXX-robust-api/README.md
waxlamp: I think that's everything.

draft/XXXX-robust-api/README.md
JackWilb: > My proposal was actually not to have the client send the finalize request, but rather, have the server send the request. All the client would need to do is to send the appropriate Etags on upload part completion.

Ah, got it. Thanks for the clarification

draft/XXXX-robust-api/README.md
JackWilb: Make sense

draft/XXXX-robust-api/README.md
jjnesbitt:
```suggestion
# RFC XXXX: Robust Upload System
```

draft/XXXX-robust-api/README.md
jjnesbitt: I agree, we should discuss this in a separate issue.

waxlamp (approved):

JackWilb (commented):
I'm 99% okay with this RFC. We should clean up the section regarding whether to store the ETags on the client or the server. I believe server was the preferred choice -- to allow resuming uploads -- but correct me if I'm wrong. If so, that'd mean removing the references to the client having to send all the ETags on finalize, too.

JackWilb (approved):
Thanks for this work, Jake!

draft/XXXX-robust-api/README.md
waxlamp: See #26.

PR #23, opened 2021-01-26
RFC: Eliminate directory structure

waxlamp:


JackWilb (approved):

jjnesbitt:
These changes seem reasonable to me, although it was my understanding that the main reasons we had this structure in the first place was to allow for easy browsing of RFCs in this repository. Under this new structure, if someone was looking for a specific RFC, they could stumble across RFCs that have been superseded, withdrawn RFCs, etc. Do we no longer care about this ease of use?

waxlamp:
> These changes seem reasonable to me, although it was my understanding that the main reasons we had this structure in the first place was to allow for easy browsing of RFCs in this repository. Under this new structure, if someone was looking for a specific RFC, they could stumble across RFCs that have been superseded, withdrawn RFCs, etc. Do we no longer care about this ease of use?

This was mainly motivated by the fact that I myself found it difficult to use the directory structure the way it is now. Then I imagined asking a new person to read this repository to get a sense for the history of the project, and it feels to me like this branching directory system would only hinder that style of browsing. Add to that the fact that I am not sure how "accepted" and "final" are different, together with the fact that as these move around, URLs elsewhere will become invalidated, and I just figured that a simpler, more straightforward, more visible structure (like that used by the Python community) would be better.

jjnesbitt (approved):

PR #24, opened 2021-01-26
RFC: PUT endpoint for appending/modifying table data

### LINKED ISSUE #19: Write table append API RFC
See https://github.com/multinet-app/multinet-server/issues/483.

waxlamp:
Closes #19

### DIFF
--- a/draft/XXXX-put-row-endpoint/README.md
+++ b/draft/XXXX-put-row-endpoint/README.md
@@ -48,7 +48,7 @@ the `PUT` endpoint to append these tranches to reconstruct the whole table.
 We are proposing to add a new endpoint:
 
 ```
-PUT /api/workspaces/{workspace}/tables/{table}
+PUT /api/workspaces/{workspace}/tables/{table}/rows
 ```
 
 the body of which will contain data formatted as "JSON lines". For each such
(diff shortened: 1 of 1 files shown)


draft/XXXX-put-row-endpoint/README.md
```diff
@@ -0,0 +1,97 @@
...
+the `PUT` endpoint to append these tranches to reconstruct the whole table.
+
+## Proposals
+
+### `PUT` endpoint
+
+We are proposing to add a new endpoint:
+
+```
+PUT /api/workspaces/{workspace}/tables/{table}
```
jjnesbitt: If the other endpoints (including your structure change section) use the `rows` suffix, shouldn't this as well?


```suggestion
PUT /api/workspaces/{workspace}/tables/{table}/rows
```

jjnesbitt:
This looks good and concise!

draft/XXXX-put-row-endpoint/README.md
waxlamp: I'm usually a proponent of smaller changes done in sequence, but you are right, if we don't make your suggested change, we'll end up with an inconsistent API until we apply the fix in the structure section.

So what the hell, let's go ahead and make that change now.

JackWilb (approved):
LGTM. Thanks for this, Roni!

jjnesbitt (approved):

PR #27, opened 2021-03-12
RFC: Add getting user permissions RFC draft

JackWilb:


draft/xxxx-getting-user-permissions/README.md
```diff
@@ -0,0 +1,48 @@
...
+in case we ever have a vulnerability allows reading from any valid workspace name.
+Again, I'm open to feedback on whether this extra step is worth the time to implement.
+
+## Backwards Compatibility
+
+Since this is a new addition, there should be no breaking changes or incompatibilities.
+
+## Reference Implementation
+
+N/A.
```
waxlamp:
```suggestion
```

draft/xxxx-getting-user-permissions/README.md
```diff
@@ -0,0 +1,48 @@
...
+endpoint to allow an application to check if the user has permissions to write to/
+read from the workspace without exposing the list of others who have access.
+
+## Proposals
+
+To satisfy the requirements of this feature, I can conceive of two ways that the work
+might be done. 
+
+1. Modify the existing permissions endpoint, `/api/workspaces/{workspace}/permissions?user=<user>`, to accept a query string parameter to filter the permissions to just the information pertinent to that user.
+2. Add a new endpoint, `/api/user/permission?workspace=<workspace>`, that returns just that users permissions for the workspace.
```
waxlamp: I think (2) is the right way to go. The semantics should be the following:

1. The REST method is `GET`.
2. If the workspace is public and the user does not have specific privileges for that workspace, then the endpoint returns `reader`.
3. If the workspace is not public and the user does not have specific privileges for that workspace, the endpoint returns `404 Workspace Not Found`.
4. If the user has reader, writer, maintainer, or owner privileges, the endpoint returns `reader`, `writer`, `maintainer`, or `owner` as appropriate.

The rationale for a separate endpoint is to avoid giving two separate semantic behaviors to the existing endpoint. That one is structured as "give me all user permissions on a specific workspace" while this new one is "tell me what privileges I have on a given workspace". They are different enough operations to merit different endpoints.

JackWilb: Sounds good. I'll update this section with the relevant info from your message

JackWilb: See bbe8f60

waxlamp (approved):

jjnesbitt (approved):
LGTM

PR #28, opened 2021-04-26
Execute RFC 0014

waxlamp:
This PR just eliminates the directory structure in favor of all RFC docs appearing the top-level.


JackWilb (changes requested):
It looks like you're missing some of the proposed changes to the template and the status metadata:

From 0014:
>1. While RFCs would still begin life in a dedicated `drafts/` folder, once
   accepted they will simply be copied to the top-level.
>2. The top-level `template.md` file can be moved to `drafts/`.
>3. The folders `accepted/`, `final/`, `rejected/`, `superseded/`, and
   `withdrawn` would disappear. `draft/` would be retained for housing draft
   RFCs as detailed in point 1 above.
>4. All existing RFCs will gain a `Status` metadata field which will inherit the
   disappeared folder name under which they used to live.
>5. The RFC template will gain a metadata field `Status`, set to `draft` by
   default, but that must be updated prior to acceptance to the appropriate
   status.

waxlamp:
@JackWilb thanks for pointing this out.

I am going to fix 4 and 5 now, and I am actually going to change the text to reflect what I did here rather than what is mentioned in 1, 2, and 3.

I will follow up in the Slack channel for details.

jjnesbitt (changes requested):
~~It seems `template.md` still resides at the top level, whereas per the RFC it should be in `drafts`~~ Moot after ee2cea4.

jjnesbitt (approved):

JackWilb (approved):

PR #31, opened 2021-05-05
RFC: New MultinetJS structure

jjnesbitt:


waxlamp (approved):
Looks great. After the current spate of PRs is settled, I think you should revisit this with design details mapping the current set of functions to method calls in one of these classes.

JackWilb (approved):
LGTM

